PR #1, opened 2017-09-15
Support for WITH, INCLUDE, OPTIONAL and fixed pagination domain

umbreak:


wwajerowicz:
retest this please

umbreak:
retest this please

umbreak:
retest this please

bogdanromanx:
I think the failures are caused by the parallel test executions with Jena code. Setting the `parallelExecution in Test := false` to all modules that use jena seems to fix it, just like in #3.

bogdanromanx:
retest this please

bogdanromanx:
The tests fail indeed because of Jena initialization; I will merge this as is and correct the build separately.

PR #3, opened 2017-09-15
Implemented filtering and exposed instance direct link navigation

### LINKED ISSUE #2: Implement instance listing filtering

bogdanromanx:
Fixes #2.


modules/indexing/src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/filtering/Op.scala
```diff
@@ -0,0 +1,64 @@
...
+  final case object Not extends LogicalOp
+  final case object Xor extends LogicalOp
+
+  object LogicalOp {
+    /**
+      * Attempts to construct ''LogicalOp'' from the provided string.
+      *
+      * @param string the string representation of the logical operator
+      */
+    final def fromString(string: String): Option[LogicalOp] = string match {
```
wwajerowicz: wouldnt `Try[LogicalOp]` as return type make more sense here? it's definitely an error when an incorrect operator is supplied

umbreak:
True. After we merge both the query builder stuff and the filtering AST we
can have one commit where we create a submodule for search.

modules/indexing/src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/filtering/Op.scala
bogdanromanx: In this particular case there's no additional information that an exception can provide. If the match succeeds the function will return the value, or return no value otherwise.

bogdanromanx:
> Not strictly related to this PR, but could we perhaps move the query/filter logic to a separate module? Queries and filters are not really used while indexing.

Agreed.

bogdanromanx:
retest this please

modules/indexing/src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/query/builder/FilteredQuery.scala
```diff
@@ -0,0 +1,246 @@
...
+  */
+object FilteredQuery {
+
+  /**
+    * Constructs a SPARQL query based on the provided filter and pagination settings that also computes the total
+    * number of results.  The filter is applied on the subjects.
+    *
+    * @param filter     the filter on which the query is based on
+    * @param pagination the pagination settings for the generated query
+    * @return a SPARQL query based on the provided filter and pagination settings
```
umbreak: This is not a SPARQL query. It contains WITH clause and INCLUDE expression

bogdanromanx:
retest this please

modules/indexing/src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/query/builder/FilteredQuery.scala
```diff
@@ -0,0 +1,246 @@
...
+    def compExpr(expr: ComparisonExpr): Stmt = {
+      val idx = nextIdx()
+      val variable = s"?${varPrefix}_$idx"
+      val ComparisonExpr(op, path, term) = expr
+      Stmt(s"$subjectVar ${path.show} $variable .", s"$variable ${op.show} ${term.show}", variable)
+    }
+
+    def fromStmts(op: LogicalOp, stmts: Vector[Stmt]): String = op match {
+      case And =>
+        val select = stmts.map(_.stmt).mkString("\n")
```
umbreak: name of the variable confusing. in SPARQL context, SELECT is a clause.
What abut ""triples"   or "triplesPattern"

modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/routes/InstanceRoutes.scala
```diff
@@ -67,15 +74,17 @@ class InstanceRoutes(
...
+                buildResponse(instanceQueries.list(domainId, filter, Pagination(from, size)), Pagination(from, size))
               }
             }
           } ~
           pathPrefix(Segment) { name =>
             pathEndOrSingleSlash {
               (get & parameter('from.as[Int] ? pagination.from) & parameter('size.as[Int] ? pagination.size) & parameter('deprecated.as[Boolean].?)) { (from, size, deprecated) =>
                 traceName("listInstancesOfSchemaName") {
-                  queryBuilder.listingQuery(domainId, name, deprecated, None, Pagination(from, size)).response
+                  val filter = Filter(deprecatedOrNoop(deprecated))
```
umbreak: extract this case class construction out of the method, so the construction of the case class is not duplicated.

bogdanromanx:
Comments to be address in a subsequent pull request.

PR #6, opened 2017-09-19
Added basic test cases for filtering.

umbreak:
Now using Blazegraph embedded instance, instead of mocked HTTP
client.


umbreak:
retest this please

PR #21, opened 2017-09-20
Documentation improvements

### LINKED ISSUE #17: Update the KG documentation to describe filters

bogdanromanx:
Changes:

- ~~add a service overview~~
- ~~describe the main service resources~~
- ~~describe the common operations on resources~~
- ~~add filtering documentation~~

Fixes #17.


docs/src/main/paradox/api-reference/operating-on-resources.md
```diff
@@ -0,0 +1,132 @@
+# Operating on resources
+
+All resources in the system share a base set of operations.  Assuming a nexus deployment at
+`http(s)://nexus.example.com` resource address of `/v0/{address}` the following operations should apply to most (all)
```
wwajerowicz: I think `path` would be more appropriate that `address`, for me `address` would be the full URL e.g. `http(s)://nexus.example.com/v0/some/resource`

bogdanromanx: When we describe RESTful resources it's common to use address instead of path. In this particular context I believe that address is more appropriate.

PR #23, opened 2017-09-20
Update to newest version of sbt-nexus.

wwajerowicz:


bogdanromanx:
The failure is caused by an unstable test.

PR #76, opened 2017-10-24
Added the possibility to have colons on the filter 'value'

### LINKED ISSUE #75: Allow filter values to contain colon
Right now the values on the 'value' property of filters does not allow colon.
Allow it in order to support filters like:

```json
{
  "@context": {
    "prov": "http://www.w3.org/ns/prov#"
  },
  "filter": {
    "op": "eq",
    "path": "prov:startedAtTime",
    "value": "2017-10-07T16:00:00-05:00"
  }
}
```

umbreak:
Fixes #75


umbreak:
retest this please

PR #82, opened 2017-10-29
Expand contexts during schema and instance validation

bogdanromanx:


modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/routes/InstanceRoutes.scala
```diff
@@ -129,57 +128,55 @@ class InstanceRoutes(instances: Instances[Future, Source[ByteString, Any], Sourc
...
+            (delete & parameter('rev.as[Long]) & authorizeResource(instanceId, Write)) { rev =>
+              traceName("deprecateInstance") {
+                onSuccess(instances.deprecate(instanceId, rev)) { ref =>
                   complete(StatusCodes.OK -> ref)
                 }
               }
+            }
+        } ~
+          path("attachment") {
+            val resource = Path(s"$instanceId/attachment")
```
umbreak: `Path(s"${instanceId.show}/attachment")`

PR #92, opened 2017-10-31
Added the meta information for each event into the primary store

### LINKED ISSUE #90: Store on the primary store the meta information for each event
Store information about the `Identity` and the `Instance` for each `Event`

umbreak:
Fixes #90

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/CallerCtx.scala
+++ b/modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/CallerCtx.scala
@@ -1,0 +1,42 @@
+package ch.epfl.bluebrain.nexus.kg.core
+
+import java.time.Clock
+
+import ch.epfl.bluebrain.nexus.commons.iam.acls.Meta
+import ch.epfl.bluebrain.nexus.commons.iam.identity.Identity.{AuthenticatedRef, UserRef}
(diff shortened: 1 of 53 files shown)


umbreak:
retest this please

umbreak:
retest this please

modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/CallerCtx.scala
```diff
@@ -0,0 +1,42 @@
...
+  * @param caller the author of the operation call
+  */
+final case class CallerCtx(clock: Clock, caller: Caller) {
+  lazy val meta = Meta(identity(), clock.instant())
+
+  private def identity(): Identity =
+    caller.identities.collectFirst {
+      case id: UserRef => id
+    } orElse (caller.identities.collectFirst {
+      case id: AuthenticatedRef => id
```
wwajerowicz: shouldn't authenticated user always have `UserRef`?

umbreak: Whenever `UserRef` is present, we prefer it. If it isn't present, I try to find `AuthenticatedRef` and if it isn't present either, I set the Identity to `Anonymous`.

Now I know that by the implementation on IAM, we would never get `AuthenticatedRef` without a `UserRef`. However, that's just the implementation on IAM, which might change.

So I think that behaviour is still correct.

PR #93, opened 2017-11-01
Merged all namespaces into one

### LINKED ISSUE #9: Unify indexed data into a single namespace
Instances, schemas, domains and organizations are now indexed in different blazegraph namespaces which makes it difficult to query across different resource types.

Resources should ideally be indexed in a single namespace using named graphs.

umbreak:
Fixes #9


umbreak:
retest this please

PR #94, opened 2017-11-02
Expand context when indexing  instances and schemas

### LINKED ISSUE #91: Expand context when indexing schemas and instances

wwajerowicz:
Fixes #91


modules/indexing/src/test/scala/ch/epfl/bluebrain/nexus/kg/indexing/instances/InstanceIndexerSpec.scala
```diff
@@ -143,8 +138,12 @@ class InstanceIndexerSpec(blazegraphPort: Int)
   }
 
   "An InstanceIndexer" should {
```
umbreak: Could you further add tests which check the whole triples, not only the ones added by the nexus namespace? Like for example in [this line](https://github.com/wwajerowicz/nexus-kg/blob/0b4f33a11968a1518e5534c61000e60504332f8c/modules/indexing/src/test/scala/ch/epfl/bluebrain/nexus/kg/indexing/instances/InstanceIndexerSpec.scala#L177) or [this other](https://github.com/wwajerowicz/nexus-kg/blob/0b4f33a11968a1518e5534c61000e60504332f8c/modules/indexing/src/test/scala/ch/epfl/bluebrain/nexus/kg/indexing/instances/InstanceIndexerSpec.scala#L189). The number of triples generated is 14 but we only verify the content of 9.

Same of `OrganizationIndexerSpec` and `SchemaIndexerSpec`

PR #101, opened 2017-11-10
Added a query parameter on listing in order to return full entities

### LINKED ISSUE #44: Instance properties are not retrieved as part of a search result

### LINKED ISSUE #100: Added HATEOAS links when fetching specific entities
The `links` field in the response's payload is only returned on response to search endpoints. This should be also return when hitting a specific resource like `/organizations/nexus` or `/domains/nexus/bbp`

umbreak:
Fixes #44
FIxes #100

### DIFF
--- a/modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/routes/SearchResponse.scala
+++ b/modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/routes/SearchResponse.scala
@@ -3,39 +3,67 @@ package ch.epfl.bluebrain.nexus.kg.service.routes
 import akka.http.scaladsl.model.{StatusCodes, Uri}
 import akka.http.scaladsl.server.Directives.{complete, extract, onSuccess}
 import akka.http.scaladsl.server.Route
+import cats.syntax.functor._
 import ch.epfl.bluebrain.nexus.commons.sparql.client.SparqlCirceSupport._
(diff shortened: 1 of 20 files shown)


modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/routes/SearchResponse.scala
```diff
@@ -3,39 +3,67 @@ package ch.epfl.bluebrain.nexus.kg.service.routes
...
-                                                         R: Encoder[UnscoredQueryResult[A]],
-                                                         S: Encoder[ScoredQueryResult[A]]): Route =
-      extract(_.request.uri) { uri =>
-        onSuccess(qr) { result =>
-          val lqu = base.copy(path = uri.path, fragment = uri.fragment, rawQueryString = uri.rawQueryString)
-          complete(StatusCodes.OK -> LinksQueryResults(result, pagination, lqu))
-        }
+    def buildResponse[Entity](fields: Set[String], base: Uri, pagination: Pagination)(
+        implicit
+        f: Id => Future[Option[Entity]],
```
wwajerowicz: does this function need to be implicit?

umbreak: I think it is better, since it is a kind of mapping. We could create those implicits conversions directly in the `Main` however, or in the routes like it is right now.

But it is not strictly needed

Issue #26, opened 2017-09-21
Index ACLs into Blazegraph


bogdanromanx:

umbreak:
We will start indexing all the Events coming from Kafka which contains `read` permissions into blazegraph. We will use the predicate `nx:read` and the value  will be the value contained on the Event's @id field. We will have as many triples as identities we have.
Those triples are stored in an own named graph (in order to perform updates to the graph easily) and have as a `subject` the resource where they belong (if there is a permission event for organizations, the subject will be the id of the organization, etc).

PR #102, opened 2017-11-13
Added indexing for ACL events

umbreak:


modules/indexing/src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/acls/AclIndexer.scala
```diff
@@ -27,26 +40,62 @@ class AclIndexer[F[_]](client: SparqlClient[F], settings: AclIndexingSettings)(i
     * @return a Unit value in the ''F[_]'' context
     */
   final def apply(event: Event): F[Unit] = event match {
-    case PermissionsCreated(path, acl, meta) =>
-      acl.acl.map { case AccessControl(id, perms) if perms(Permission.Read) => add(path, id) }
-      F.pure(())
-    case PermissionsAdded(path, identity, permissions, meta) if permissions(Permission.Read) =>
-      add(path, identity)
-    case PermissionsSubtracted(path, identity, permissions, meta) if permissions(Permission.Read) =>
+    case PermissionsCreated(path, acl, _) =>
```
wwajerowicz: don't we meed `&& isKg(path)` here as well?

umbreak: yes. I forgot!

modules/indexing/src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/acls/AclIndexer.scala
```diff
@@ -27,26 +40,62 @@ class AclIndexer[F[_]](client: SparqlClient[F], settings: AclIndexingSettings)(i
...
-    case PermissionsRemoved(path, identity, meta) =>
+    case PermissionsRemoved(path, identity, _) if isKg(path) =>
       remove(path, identity)
     case _ => F.pure(())
   }
 
-  // TODO: implement
-  private def add(path: Path, identity: Identity): F[Unit] = ???
+  private def isKg(path: Path): Boolean =
+    Path(s"${path.reverse.head}") == Path("kg")
```
hygt: Wouldn't `path.reverse.head.toString == "kg"` work the same?

umbreak: If the path in the `Event` is just `/kg` then it would be `"/kg" == "kg"`
So it is to double check special cases with slashes

umbreak:
Fixes #26

PR #108, opened 2017-11-15
Added an upper limit for size query parameter on search endpoints

### LINKED ISSUE #30: Cap pagination query parameters at reasonable values

umbreak:
Fixes #30

### DIFF
--- a/modules/service/src/main/resources/sparql.conf
+++ b/modules/service/src/main/resources/sparql.conf
@@ -44,9 +44,9 @@ app {
 
     # Default pagination information
     page-from = 0
-    page-size = 20
+    page-size = 10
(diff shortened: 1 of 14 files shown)


modules/service/src/main/resources/sparql.conf
```diff
@@ -45,5 +45,8 @@ app {
     # Default pagination information
     page-from = 0
     page-size = 20
+    page-size = ${?DEAFULT_SEARCH_SIZE}
```
bogdanromanx: typo

I would also suggest renaming to: `PAGINATION_DEFAULT_SIZE` and `PAGINATION_MAX_SIZE` respectively.

PR #116, opened 2017-11-17
Wrap kafka consumer in a BackoffSupervisor

### LINKED ISSUE #112: Use cluster singleton for Kafka Consumer

hygt:
Fixes #112

### DIFF
--- a/modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/queue/KafkaConsumer.scala
+++ b/modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/queue/KafkaConsumer.scala
@@ -1,49 +1,63 @@
 package ch.epfl.bluebrain.nexus.kg.service.queue
 
 import akka.Done
-import akka.actor.ActorSystem
+import akka.actor._
 import akka.kafka.ConsumerMessage.CommittableOffsetBatch
 import akka.kafka.scaladsl.Consumer
 import akka.kafka.{ConsumerSettings, Subscriptions}
+import akka.pattern.BackoffSupervisor.{CurrentChild, GetCurrentChild}
+import akka.pattern.{Backoff, BackoffSupervisor, ask}
 import akka.stream.ActorMaterializer
 import akka.stream.scaladsl.Sink
-import io.circe.Decoder
+import akka.util.Timeout
+import com.google.common.annotations.VisibleForTesting
+import io.circe._
 import io.circe.parser._
 import journal.Logger
 
+import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
 
-object KafkaConsumer {
+class KafkaConsumerWrapper[T](settings: ConsumerSettings[String, String],
+                       index: T => Future[Unit],
+                       topic: String,
+                       decoder: Decoder[T])
+    extends Actor {
+
+  private implicit val ec: ExecutionContext  = context.dispatcher
+  private implicit val to: Timeout           = 30.seconds
+  private implicit val mt: ActorMaterializer = ActorMaterializer()
+  private implicit val ed: Decoder[T]        = decoder
 
   private val log = Logger[this.type]
 
-  /**
-    * Starts a Kafka consumer that reads messages from a particular ''topic'', decodes them into
-    * events of type ''T'' and indexes them using the provided function.
-    *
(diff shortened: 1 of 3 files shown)


modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/queue/KafkaConsumer.scala
```diff
@@ -1,56 +1,134 @@
...
+    supervisor
+  }
+
+  /**
+    * Kills the managed actor that wraps the Kafka consumer, which should be re-spawned by its supervisor.
+    *
+    * @param supervisor the supervisor actor handle
+    */
+  @VisibleForTesting
+  private[queue] def stop(supervisor: ActorRef)(implicit as: ActorSystem): Unit = {
```
umbreak: This method should be on the tests if it is used only when testing, right?
Besides that, `scapegoat` is complaining because of the use of `ref.get`

modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/queue/KafkaConsumer.scala
```diff
@@ -1,56 +1,134 @@
...
-      implicit as: ActorSystem,
-      D: Decoder[T]): Future[Done] = {
-    implicit val ec: ExecutionContext  = as.dispatcher
-    implicit val mt: ActorMaterializer = ActorMaterializer()
+  override def receive: Receive = {
+    case value: String =>
+      decode[T](value) match {
+        case Right(event) =>
+          log.debug(s"Received message: $value")
+          index(event).onComplete {
```
umbreak: I would try to retry whenever the index function fails, and if it keeps failing after retry I would store it somewhere. You can probably copy the logic from `SequentialTagIndexer` since we do the same: https://github.com/BlueBrain/nexus-commons/blob/26e9e5e19ef3331554aabd87c14ab95940df8b01/modules/service/src/main/scala/ch/epfl/bluebrain/nexus/commons/service/persistence/SequentialTagIndexer.scala#L84

modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/queue/KafkaConsumer.scala
```diff
@@ -1,56 +1,134 @@
...
+            case Success(_) => ()
+            case Failure(e) => log.error(s"Failed to index event: $event; skipping.", e)
+          }
+          sender ! Done
```
bogdanromanx: I think `Done` should only be sent once the indexing function completes otherwise there's no backpressure and since the mailbox is unbounded it can result and out of memory errors.

modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/queue/KafkaConsumer.scala
```diff
@@ -1,56 +1,134 @@
...
+        case Left(e) =>
+          log.error(s"Failed to decode message: $value; skipping.", e)
+          sender ! Done
+        // $COVERAGE-ON$
+      }
+  }
 
-    Consumer
+  override def preStart(): Unit = {
+    val done = Consumer
```
bogdanromanx: could we embed a killswitch and trip it when failures are detected instead of exchanging messages with the underlying actor?

modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/queue/KafkaConsumer.scala
```diff
@@ -123,12 +145,10 @@ object KafkaConsumer {
     *
     * @param supervisor the supervisor actor handle
     */
-  @VisibleForTesting
   private[queue] def stop(supervisor: ActorRef)(implicit as: ActorSystem): Unit = {
     implicit val ec: ExecutionContext = as.dispatcher
     implicit val to: Timeout          = 30.seconds
-    val child                         = supervisor ? GetCurrentChild
-    child.foreach(_.asInstanceOf[CurrentChild].ref.get ! PoisonPill)
+    (supervisor ? GetCurrentChild).mapTo[KafkaConsumerWrapper[_]].foreach(_.kill())
```
bogdanromanx: calling functions on the underlying actor breaks encapsulation, use messages instead even if this is only used in testing.

hygt:
Jenkins retest this please

PR #119, opened 2017-11-20
Added capability of fetching specific revision when fetching resources

### LINKED ISSUE #97: Specific revision cannot be retrieved from organizations / domains / schemas / context
Only `instances` allow us to use the query parameter `rev` to return a concrete revision.
We have to extend this to the other resources in the system

umbreak:
Fixes #97

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/contexts/Contexts.scala
+++ b/modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/contexts/Contexts.scala
@@ -220,21 +220,16 @@ final class Contexts[F[_]](agg: ContextAggregate[F], doms: Domains[F], baseUri: 
     *         ''F[_]'' otherwise
     */
   def fetch(id: ContextId, rev: Long): F[Option[Context]] =
-    fetchCurrent(id, rev).map {
-      case None    => None
-      case Some(c) => Some(Context(c.id, rev, c.value, c.deprecated, c.published))
+    stateAt(id, rev).map {
+      case c: Current if c.rev == rev => Some(Context(c.id, rev, c.value, c.deprecated, c.published))
+      case _                          => None
     }
 
-  private def fetchCurrent(id: ContextId, rev: Long): F[Option[Current]] =
-    agg
-      .foldLeft[ContextState](id.show, Initial) {
-        case (state, ev) if ev.rev <= rev => Contexts.next(state, ev)
-        case (state, _)                   => state
-      }
-      .map {
-        case c: Current if c.rev == rev => Some(c)
-        case _                          => None
-      }
+  private def stateAt(id: ContextId, rev: Long): F[ContextState] =
+    agg.foldLeft[ContextState](id.show, Initial) {
+      case (state, ev) if ev.rev <= rev => Contexts.next(state, ev)
+      case (state, _)                   => state
+    }
 
   /**
     * Expands the argument context representation by recursively loading referenced contexts regardless of their status.
--- a/modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/contexts/Contexts.scala
+++ b/modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/contexts/Contexts.scala
@@ -209,6 +209,34 @@ final class Contexts[F[_]](agg: ContextAggregate[F], doms: Domains[F], baseUri: 
     }
 
   /**
+    * Queries the system for the context identified by the argument id at the specified revision.  The (in)existence of
+    * the context or the requested revision is represented by the [[scala.Option]] type wrapped within the ''F[_]''
+    * context.
+    *
+    * @param id  the unique identifier of the context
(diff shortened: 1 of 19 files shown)


modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/contexts/Contexts.scala
```diff
@@ -208,6 +208,34 @@ final class Contexts[F[_]](agg: ContextAggregate[F], doms: Domains[F], baseUri:
...
+    */
+  def fetch(id: ContextId, rev: Long): F[Option[Context]] =
+    fetchCurrent(id, rev).map {
+      case None    => None
+      case Some(c) => Some(Context(c.id, rev, c.value, c.deprecated, c.published))
+    }
+
+  private def fetchCurrent(id: ContextId, rev: Long): F[Option[Current]] =
+    agg
+      .foldLeft[ContextState](id.show, Initial) {
```
bogdanromanx: Stitch together the fold and the map to avoid the extra scheduling of the transformation at the end.

bogdanromanx: it also makes it safer as you don't need the second pattern match in the transformation.

modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/domains/Domains.scala
```diff
@@ -104,6 +104,34 @@ final class Domains[F[_]](agg: DomainAggregate[F], orgs: Organizations[F])(impli
...
+    */
+  def fetch(id: DomainId, rev: Long): F[Option[Domain]] =
+    fetchCurrent(id, rev).map {
+      case None    => None
+      case Some(c) => Some(Domain(c.id, rev, c.deprecated, c.description))
+    }
+
+  private def fetchCurrent(id: DomainId, rev: Long): F[Option[Current]] =
+    agg
+      .foldLeft[DomainState](id.show, Initial) {
```
bogdanromanx: same as for contexts

modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/organizations/Organizations.scala
```diff
@@ -93,6 +93,34 @@ final class Organizations[F[_]](agg: OrgAggregate[F])(implicit F: MonadError[F,
...
+    */
+  def fetch(id: OrgId, rev: Long): F[Option[Organization]] =
+    fetchCurrent(id, rev).map {
+      case None    => None
+      case Some(c) => Some(Organization(c.id, rev, c.value, c.deprecated))
+    }
+
+  private def fetchCurrent(id: OrgId, rev: Long): F[Option[Current]] =
+    agg
+      .foldLeft[OrgState](id.show, Initial) {
```
bogdanromanx: same as for contexts

modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/schemas/Schemas.scala
```diff
@@ -180,13 +180,41 @@ final class Schemas[F[_]](agg: SchemaAggregate[F], doms: Domains[F], ctxs: Conte
...
+    */
+  def fetch(id: SchemaId, rev: Long): F[Option[Schema]] =
+    fetchCurrent(id, rev).map {
+      case None    => None
+      case Some(c) => Some(Schema(c.id, rev, c.value, c.deprecated, c.published))
+    }
+
+  private def fetchCurrent(id: SchemaId, rev: Long): F[Option[Current]] =
+    agg
+      .foldLeft[SchemaState](id.show, Initial) {
```
bogdanromanx: same as for contexts

modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/routes/DomainRoutes.scala
```diff
@@ -74,10 +74,13 @@ final class DomainRoutes(domains: Domains[Future], domainQueries: FilterQueries[
   protected def readRoutes(implicit credentials: Option[OAuth2BearerToken]): Route =
     (extractDomainId & pathEndOrSingleSlash) { domainId =>
       (get & authorizeResource(domainId, Read)) {
-        traceName("getDomain") {
-          onSuccess(domains.fetch(domainId)) {
-            case Some(domain) => complete(domain)
-            case None         => complete(StatusCodes.NotFound)
+        parameter('rev.as[Long].?) { revOpt =>
```
bogdanromanx: This style of handling revision does make the code shorter but maybe a little more difficult to read and modify.

Could we consider the following style instead?
```
parameter('rev.as[Long].?) {
  case Some(rev) =>
    traceName("getDomainRevision") {
      onSuccess(...) {}
    }
  case None =>
    traceName("getDomain") {
      onSuccess(...) {}
    }
}
```

PR #145, opened 2017-12-01
Fixed the addition of the links as triples in Blazegraph

### LINKED ISSUE #142: When updating instances from response, links is added to the indexing strategy
If you want to update an instance into the platform and you use the representation of the previous instance for it (fetching the previous revision from the platform), `links` gets indexed into Blazegraph, while they shouldn't.

In general, any field which is introduced by the platform and it is not part of the origin payload of the client should not be indexed into Blazegraph (or if it has to be indexed, its value shouldn't be taken by the clients payload).

umbreak:
Fixes #142


modules/indexing/src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/JsonManipulation.scala
```diff
@@ -0,0 +1,29 @@
...
+    */
+  implicit class JsonManipulationSyntax(json: Json) {
+
+    /**
+      * Method exposed on Json instances.
+      *
+      * @param keys list of ''keys'' to be removed from the top level of the ''json''
+      * @return the original json without the provided ''keys'' on the top level of the structure
+      */
+    def removeKeys(keys: String*): Json = {
```
bogdanromanx: I think this should go in nexus-commons along with the key sorting under a JsonOps type.

umbreak: ok!

PR #157, opened 2017-12-11
Add search and distribution contexts to responses

hygt:
- Add search context
- Move attachment metadata to 'distribution' field in response, with its own context


hygt:
Fixes #140 #141

PR #165, opened 2017-12-13
Update keys according to their definition in distribution context

### LINKED ISSUE #162: Attachment fields do not match the ones defined on @context
Fields defined in the context https://bbp-nexus.epfl.ch/dev/v0/contexts/nexus/core/distribution/v0.1.0 do not match the fields on the actual response.

We should change those fields:
`contentSize` instead of `size`
`mediaType` instead of `contentType`
`algorithm` instead of `alg`

hygt:
Fixes #162


hygt:
Jenkins retest this please

PR #184, opened 2018-01-04
 Added sort query param for search endpoints

### LINKED ISSUE #57: Add `sort` query param to describe the sorting criteria when querying data
the query API [](https://bbp-nexus.epfl.ch/dev/docs/kg/api-reference/instances.html#full-text-search) allows to define a filter.
However one cannot specify how the results are sorted.
Let's say I want to display my model sorted by alphabetic order of their name. I cannot do that if there is pagination involved.

umbreak:
Fixes #57

### DIFF
--- a/modules/indexing/src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/query/Sort.scala
+++ b/modules/indexing/src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/query/Sort.scala
@@ -17,24 +17,26 @@ final case class SortList(values: List[Sort]) extends Product with Serializable 
   /**
     * @return the string representation of the variables used for sorting (inside the SELECT clause) in SPARQL language
     */
-  def toVars = toVarsMapping map {case (_, variable) => variable} mkString (" ")
+  def toVars = toVarsMapping.map { case (_, variable) => variable }.mkString(" ")
 
   /**
     * @return the string representation of the optional triples used for sorting (inside the WHERE clause) in SPARQL language
     */
   def toTriples =
-    toVarsMapping map {
-      case (Sort(_, predicate), variable) => s"?s <${predicate}> $variable"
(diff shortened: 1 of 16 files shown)


modules/indexing/src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/query/Sort.scala
```diff
@@ -0,0 +1,93 @@
...
+  * Data type of a collection of [[Sort]]
+  * @param values the values to be sorted in the provided order
+  */
+final case class SortList(values: List[Sort]) extends Product with Serializable {
+  private lazy val toVarsMapping = values.zipWithIndex.map { case (sort, i) => sort -> s"?sort$i" }
+
+  /**
+    * @return the string representation of the variables used for sorting (inside the SELECT clause) in SPARQL language
+    */
+  def toVars = toVarsMapping map {case (_, variable) => variable} mkString (" ")
```
hygt: Formatting.

modules/indexing/src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/query/Sort.scala
```diff
@@ -0,0 +1,93 @@
...
+
+  /**
+    * @return the string representation of the optional triples used for sorting (inside the WHERE clause) in SPARQL language
+    */
+  def toTriples =
+    toVarsMapping map {
+      case (Sort(_, predicate), variable) => s"?s <${predicate}> $variable"
+    } mkString ("}\nOPTIONAL{") prefixSuffixNonEmpty ("OPTIONAL{", "}")
```
hygt: I think we should not overuse infix notations. For something like `mkString` it's not very natural.
http://docs.scala-lang.org/style/method-invocation.html#infix-notation

umbreak:
retest this please

Issue #98, opened 2017-11-09
Expand JSON-LD on response


umbreak:
When retrieving a resource, we would like to add the possibility to represent it as an expanded JSON-LD.

umbreak:
We were discussing yesterday how to represent that in terms of API. We thought about having a query param called `output` or `format` which has the following option

- default
- jsonld-expand
- jsonld-compact
- json

and it will be used like `...?format=jsonld-expand`

MFSY:
I like ?format=....
But you may want to look at Jena json-ld rdf formats:
https://jena.apache.org/documentation/io/rdf-output.html#json-ld

Json-ld is serialized in Jena using different flavors: "compacted", "expanded", "flattened", "framed"

umbreak:
Yes you are right. In the POC I did I used the Jena [JsonLDWriter](https://jena.apache.org/documentation/javadoc/arq/org/apache/jena/riot/writer/JsonLDWriter.html) which allows you to specify those types. So then we could have:

- compacted
- expanded
- flattened
- framed
- default (without formatting, as the customer pushed it)

PR #187, opened 2018-01-09
Used nxv:Schema as type for schemas

MFSY:


hygt:
I will update the commons version as part of #98 
There are breaking changes in the way errors get marshalled.

umbreak:
retest this please

genric:
Any estimate when this could be merged? As we would prefer to try new updates to the distribution schema ASAP. Thanks.

PR #212, opened 2018-02-01
Added context query parameter to be used by sort, field and filter

### LINKED ISSUE #209: Allow search operations to define a context query param. to be used in other query params.
Right now, the `filter` query parameter defines the @context on itself:
```json
filter={
   "@context" : {...},
   "filter" : {...}
}
```
However, we would like to have an extra query parameter called `context` which defines the context (with its prefix mappings) that can be used not only in the `filter` query parameter but also elsewhere (specifically in `sort` and `field` query parameters).

umbreak:
Also I moved all filter to commons (since it will potentially be also used in Elasticsearch) and provides a decoder for Filter which allows easier implementation of POST search entity in further
commits.

Fixes #209


umbreak:
Jenkins retest this please

umbreak:
Jenkins retest this please

umbreak:
Jenkins retest this please

PR #216, opened 2018-02-02
Enable listings in Elastic Search.

wwajerowicz:


modules/elastic/src/main/resources/es-index.json
```diff
@@ -1,6 +1,17 @@
...
+      "properties": {
+        "{{rdfType}}": {"type": "keyword"},
+        "{{contextGroup}}": {"type": "keyword"},
+        "{{organization}}": {"type": "keyword"},
+        "{{domain}}": {"type": "keyword"},
+        "{{schema}}": {"type": "keyword"},
+        "{{schemaGroup}}": {"type": "keyword"},
+        "{{deprecated}": {"type": "boolean"},
+        "{{published}}": {"type": "boolean"},
+        "@id": {"type": "keyword"}
```
umbreak: You may want to add `"@type": {"type": "keyword"}` too

modules/elastic/src/main/scala/ch/epfl/bluebrain/nexus/kg/query/BaseElasticQueries.scala
```diff
@@ -0,0 +1,98 @@
...
+    )
+  protected def term(key: String, value: String): Json =
+    Json.obj(
+      "term" -> Json.obj(
+        key -> Json.fromString(value)
+      )
+    )
+
+  private def matchAllQuery(filterTerms: Json) =
+    Json.obj(
```
umbreak: I think you don't need all of that. You can have simply `Json.obj("query" -> Json.obj("bool" -> Json.obj("filter" -> Json.obj("bool" -> Json.obj("must" -> Json.arr(terms: _*))))))`

modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/routes/ContextRoutes.scala
```diff
@@ -149,27 +159,55 @@ class ContextRoutes(contexts: Contexts[Future], contextQueries: FilterQueries[Fu
           val filter     = filterFrom(deprecatedOpt, None, querySettings.nexusVocBase) and publishedExpr(publishedOpt)
           implicit val _ = (id: ContextId) => contexts.fetch(id)
           (pathEndOrSingleSlash & authenticateCaller) { implicit caller =>
-            contextQueries.list(filter, pagination, None, sort).buildResponse(fields, base, prefixes, pagination)
+            sort match {
+              case SortList.Empty =>
```
umbreak: I would rather have a class which contains both, the `ElasticQueries ` and `SparqlQueries` which depending on the filter, sorting, pagination and so on, decides to choose the elastic indexer or the sparql indexer. The decision to use one or the other can trickier than just if sort query params are defined or not

modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/routes/ContextRoutes.scala
```diff
@@ -150,27 +160,55 @@ class ContextRoutes(contextQueries: FilterQueries[Future, ContextId], base: Uri)
             val filter     = filterFrom(deprecatedOpt, None, querySettings.nexusVocBase) and publishedExpr(publishedOpt)
             implicit val _ = (id: ContextId) => contexts.fetch(id)
             (pathEndOrSingleSlash & authenticateCaller) { implicit caller =>
-              contextQueries.list(filter, pagination, None, sort).buildResponse(fields, base, prefixes, pagination)
+              sort match {
+                case SortList.Empty =>
```
bogdanromanx: Can we sort resources based on arbitrary properties?

wwajerowicz: I think we could, but it would require the property on which we sort to match exactly what we indexed in elastic search, e.g. `nxv:rev` and `{nexusBase}/vocabs/term/v1.0.0/rev` would not be treated as the same. We could do this like that for now and later expand the properties to full URIs when indexing.

wwajerowicz: But for now we could do it for properties which we have in ES with full URIs, like `deprecated`, `updatedAt` etc.

bogdanromanx: Ok, let's avoid sorting for now and handle it in a separate pull request.

build.sbt
```diff
@@ -191,17 +192,49 @@ lazy val service = project
...
+lazy val testsElastic = project
+  .in(file("modules/tests-elastic"))
+  .dependsOn(core % "test->test;compile->compile", service % "test->test;compile->compile")
+  .settings(common)
+  .settings(
+    name := "kg-tests-elastic",
+    moduleName := "kg-tests-elastic",
+    libraryDependencies ++= Seq(
+      commonsTest                  % Test,
+      "io.netty"                   % "netty" % "3.10.6.Final" % Test,
```
bogdanromanx: pull versions to the top of the file

modules/elastic/src/main/scala/ch/epfl/bluebrain/nexus/kg/ElasticIdDecoder.scala
```diff
@@ -0,0 +1,28 @@
...
+    * Create and implicit ID decoder which will try to decode `A` from `@id` field in JSON object using
+    * `ConfiguredQualifier[A]`
+    * @param Q    ConfiguredQualifier for A
+    * @param T    Typeable for A
+    * @tparam A   type of object to be decooded
+    * @return     `Deocder` for `S`
+    */
+  final implicit def elasticIdDecoder[A](implicit Q: ConfiguredQualifier[A], T: Typeable[A]): Decoder[A] =
+    Decoder.decodeJson.emap { json =>
+      json.hcursor
```
bogdanromanx: this can probably be simplified to something like:
```
json.hcursor
  .get[String]("@id")
  .flatMap(_.unqualify[A])
  .left.map(_ => s"Couldn't decode ${T.describe} from '@id' in ${json.noSpaces}")
```

modules/elastic/src/main/scala/ch/epfl/bluebrain/nexus/kg/query/BaseElasticQueries.scala
```diff
@@ -0,0 +1,165 @@
...
+/**
+  * Base class implementing common Elastic Search queries
+  *
+  * @param elasticClient  Elastic Search client
+  * @param settings       Elastic Search settings
+  * @param rs             HTTP client
+  * @tparam F             the monadic effect type
+  * @tparam Id            type of the ID that's returned
+  */
+abstract class BaseElasticQueries[F[_], Id](elasticClient: ElasticClient[F], settings: ElasticIndexingSettings)(
```
umbreak: Is there a reason why `fields` is not used to return certain fields? I've seen that you call elasticsearch listings only when fields are `Field.Empty`.
In case of Fields.Empty we want to retrieve only the `@id`, so it would be more performant to specify this on the calls to `client.search...`

wwajerowicz: I wanted to implement the minimal feature set for this PR, and possibly add more features in following PRs.

wwajerowicz:
retest this please

wwajerowicz:
retest this please

umbreak:
Jenkins retest this please

wwajerowicz:
retest this please

umbreak:
Jenkins retest this please

umbreak:
Jenkins retest this please

umbreak:
Jenkins retest this please

umbreak:
Jenkins retest this please

PR #230, opened 2018-02-12
Modified Sparql query for ACLs check based on IAM information

### LINKED ISSUE #137: SPARQL query optimization: ACLs check only on domain level
Optimize the SPARQL queries done when hitting a search endpoint.

### LINKED ISSUE #210: Change sparql query to check on ACLs from the result obtained from IAM
We don't need to index ACLs in Blazegraph if we are going check up to the domain level.

In the same terms, we don't need to specify acls on the queries, if we already have the required information from IAM: we know if we have root access, if we have access on an organization level or if we have access to some domains. With that information we can build the necessary queries.

### LINKED ISSUE #229: Delete KafkaConsumer form KG
As explained on #210 and #137 we don't need the ACLs indexed on the triple store because we are not going to query for them in this fashion.

umbreak:
Fixes #210
Fixes #137
Fixes #229


modules/sparql/src/test/scala/ch/epfl/bluebrain/nexus/kg/indexing/BlazegraphSpecLike.scala
```diff
@@ -2,13 +2,12 @@ package ch.epfl.bluebrain.nexus.kg.indexing
```
hygt: To be cleaned up after #216 is merged?

umbreak:
Jenkins retest this please

umbreak:
Jenkins retest this please

PR #245, opened 2018-02-14
Enforce ACLs in ES listing.

wwajerowicz:


modules/elastic/src/main/scala/ch/epfl/bluebrain/nexus/kg/query/instances/InstancesElasticQueries.scala
```diff
@@ -30,16 +33,17 @@ class InstancesElasticQueries[F[_]](elasticClient: ElasticClient[F], settings: E
...
   def list(pagination: Pagination,
            schemaName: SchemaName,
            deprecated: Option[Boolean],
-           published: Option[Boolean]): F[QueryResults[InstanceId]] = {
-    elasticClient.search[InstanceId](query(termsFrom(deprecated, published) :+ schemaGroupTerm(schemaName): _*))(
-      pagination,
-      sort = defaultSort)
+           published: Option[Boolean],
+           acls: FullAccessControlList): F[QueryResults[InstanceId]] = {
+    elasticClient.search[InstanceId](query(acls, termsFrom(deprecated, published) :+ schemaGroupTerm(schemaName): _*),
```
umbreak: query is not needed here. Just check if `acls` on the path `schemaName.domainId` (with the prefix) have read permissions and execute the query for that specific index. Otherwise just don't execute it. Similar to [this method](https://github.com/BlueBrain/nexus-kg/blob/v0/modules/sparql/src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/query/builder/FilterQueries.scala#L48)

wwajerowicz: we still need query for deprecated, published as well as for sorting, but I removed some unnecessary terms in some places.
There's still some room for optimisation though.

modules/elastic/src/main/scala/ch/epfl/bluebrain/nexus/kg/query/BaseElasticQueries.scala
```diff
@@ -166,6 +166,22 @@ abstract class BaseElasticQueries[F[_], Id](elasticClient: ElasticClient[F], set
...
+      .map {
+        case (path, _) =>
+          path.segments match {
+            case `kgRoot` :: Nil                                     => true
+            case `kgRoot` :: domainId.orgId.id :: Nil                => true
+            case `kgRoot` :: domainId.orgId.id :: domainId.id :: Nil => true
+            case _                                                   => false
+          }
+      }
+      .exists(b => b)
```
umbreak: `.contains(true)` ?

PR #260, opened 2018-03-05
Added Content-Disposition header on fetch attachment

### LINKED ISSUE #258: Attachments download
I faced with an issue when I tried to download [attachment](https://bbp-nexus.epfl.ch/dev/v0/data/bbp/electrophysiology/trace/v0.1.0/01dde9fc-8504-413e-9e48-23122b9dfbec/attachment). Filename is not the proper one. Can you fix this according to originalfilename field? Many thanks

umbreak:
Fixes #258

### DIFF
--- a/modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/routes/InstanceRoutes.scala
+++ b/modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/routes/InstanceRoutes.scala
@@ -1,9 +1,10 @@
 package ch.epfl.bluebrain.nexus.kg.service.routes
 
+import java.net.URLDecoder
 import java.time.Clock
 
 import akka.http.scaladsl.model._
-import akka.http.scaladsl.model.headers.OAuth2BearerToken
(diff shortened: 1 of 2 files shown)


modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/routes/InstanceRoutes.scala
```diff
@@ -188,7 +192,10 @@ class InstanceRoutes(instances: Instances[Future, Source[ByteString, Any], Sourc
                   case Some((info, source)) =>
                     val ct =
                       ContentType.parse(info.mediaType).getOrElse(ContentTypes.`application/octet-stream`)
-                    complete(HttpEntity(ct, info.contentSize.value, source))
+                    val filename = decodedFilenameOrElse(info, "attachment")
+                    respondWithHeaders(RawHeader(s"Content-Disposition", s"attachment; filename*= UTF-8''$filename")) {
```
bogdanromanx: the filename needs to be encoded

PR #264, opened 2018-03-09
Add Kamon tracing

hygt:


modules/service/src/main/scala/ch/epfl/bluebrain/nexus/kg/service/routes/InstanceRoutes.scala
```diff
@@ -247,7 +248,7 @@ class InstanceRoutes(instances: Instances[Future, Source[ByteString, Any], Sourc
               (authenticateCaller & authorizeResource(resource, Write)) { implicit caller =>
                 fileUpload("file") {
                   case (metadata, byteSource) =>
-                    operationName("createInstanceAttachment") {
+                    tracing.trace("createInstanceAttachment") {
```
bogdanromanx: needs setting `includeRequestEntity = false` to exclude tracing the attachment regardless of its size or mediatype

PR #278, opened 2018-03-28
Added logback and codecov

umbreak:


.jvmopts
```diff
@@ -5,3 +5,4 @@
 -XX:+TieredCompilation
 -XX:+UseG1GC
 -XX:MaxMetaspaceSize=2G
+-XX:-UseGCOverheadLimit
```
hygt: I don't think this does anything with the G1 GC

umbreak:
@hygt I don't have access to merge in this project. Can you give access to me and to @wwajerowicz ?

PR #294, opened 2018-04-06
Resolved imported data and use it during validation

umbreak:


modules/core/src/test/resources/resolver/schemas/schema.json
```diff
@@ -0,0 +1,258 @@
...
+    },
+    "seeAlso": {
+      "@id": "rdfs:seeAlso",
+      "@type": "@id"
+    },
+    "this": "{{base}}/schemas/nexus/core/quantitativevalue/v0.3.0/shapes/"
+  },
+  "@type": "nxv:Schema",
+  "imports": [
+    "{{base}}/data/nexus/ontologies/unit/v0.1.0/62395ccd-bd3d-4fbb-8ca0-beb333d563b6"
```
bogdanromanx: I think this schema should also import another schema to make sure that the behaviour doesn't break in that scenario.

modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/AggregatedImportResolver.scala
```diff
@@ -20,11 +22,15 @@ class AggregatedImportResolver[F[_]](
     instanceImportResolver: InstanceImportResolver[F])(implicit F: MonadError[F, Throwable])
     extends ImportResolver[F] {
 
-  override def apply(schema: ShaclSchema): F[Set[ShaclSchema]] =
-    for {
-      setSchemas   <- schemaImportResolver(schema)
-      setInstances <- instanceImportResolver(schema)
-    } yield (setSchemas ++ setInstances)
+  override def apply(schema: ShaclSchema): F[Set[ShaclSchema]] = {
```
bogdanromanx: Do we still need the AggregatedImportResolver?

modules/core/src/test/resources/resolver/schemas/schema.json
bogdanromanx: What I mean by that is a test with a schema that imports both another schema and an instance to make sure we don't throw an error.

PR #296, opened 2018-04-12
Added commands, events, state and serialization for them

umbreak:


modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/persistence/Serializer.scala
```diff
@@ -0,0 +1,72 @@
...
+      manifest match {
+        case `EventName` =>
+          decode[Event](new String(bytes, UTF_8)) match {
+            case Left(err)    => throw err
+            case Right(value) => value
+          }
+        case _ => throw new NotSerializableException(manifest)
+      }
+
+    val identifier = 1234567
```
hygt: Is this OK? :smile:

bogdanromanx: NO! :)

How about `1050` (`"nexus-json".getBytes.map(_.toInt).sum`) ?

modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/resources/Key.scala
```diff
@@ -0,0 +1,23 @@
...
+import io.circe.{Decoder, Encoder, Json}
+
+/**
+  * The unique key which identifies a representation of a resource
+  *
+  * @param projectRef the identifier for the project. TODO: Change the type to a [ProjectReference]
+  * @param resourceId the unique identifier (inside this project) of the current resource. TODO: Change the type to a [IRI] or [Id]
+  * @param schemaId   the unique identifier (inside this project) of the schema which constrains this resource. TODO: Change the type to a [IRI] or [Id]
+  */
+final case class Key(projectRef: String, resourceId: String, schemaId: String)
```
bogdanromanx: how about `ResourceId`? `Key` seems a bit generic...

modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/resources/Payload.scala
```diff
@@ -0,0 +1,32 @@
+package ch.epfl.bluebrain.nexus.kg.core.resources
+
+import io.circe.Json
+
+/**
+  * Enumeration type for payloads.
+  */
+sealed trait Payload extends Product with Serializable
```
bogdanromanx: I'm wondering if it would be better to define this as:
`final case class Payload(mediaType: String, value: String)`
thoughts?

modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/resources/attachment/Attachment.scala
```diff
@@ -0,0 +1,48 @@
...
+// TODO: Use refined type for fileUri
+//    final def apply(uri: String, info: Info): Either[String, Attachment] =
+//      if (uri == null || uri.isEmpty) Left("'fileUri' field cannot be empty")
+//      else Right(new Attachment(uri, info))
+
+  /**
+    * Holds all metadata information related to an attachment
+    * that we want to expose through the service.
+    *
+    * @param originalFileName the original filename of the attached file
```
bogdanromanx: if we support multiple attachments, we can now consider using `filename` instead of `originalFileName`.

modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/persistence/Serializer.scala
umbreak: Yes: https://doc.akka.io/docs/akka/2.5.4/java/serialization.html

modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/resources/Key.scala
```diff
@@ -0,0 +1,23 @@
...
+object Key {
+
+  implicit val encodeKey: Encoder[Key] = Encoder.encodeJson.contramap {
+    case Key(project, resource, schema) =>
+      Json.obj("@id"     -> Json.fromString(resource),
+               "project" -> Json.fromString(project),
```
bogdanromanx: maybe we should also handle the `project` as an `@id` (in terms of jsonld) to ensure an accurate Event description in JsonLd

modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/persistence/Serializer.scala
bogdanromanx: it doesn't mean we should use `1234567`...

modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/resources/Key.scala
umbreak: the resource id is one of the fields of the key, so calling the whole thing `ResourceId` is weird

modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/resources/Payload.scala
umbreak: Why this should be better? I prefer to have a more strict type. In encoding and decoding is easier. We tried the other approach already

PR #307, opened 2018-04-19
Added some refined types

umbreak:


modules/core/src/main/scala/ch/epfl/bluebrain/nexus/kg/core/resources/attachment/Attachment.scala
```diff
@@ -3,51 +3,56 @@ package ch.epfl.bluebrain.nexus.kg.core.resources.attachment
 import java.util.UUID
 
 import ch.epfl.bluebrain.nexus.kg.core.resources.attachment.FileStream.StoredSummary
-
+import eu.timepit.refined.api.RefType.refinedRefType.unsafeWrap
+import eu.timepit.refined.api.Refined
+import eu.timepit.refined.collection.NonEmpty
+import eu.timepit.refined.string.Uuid
```
hygt: Let's add a blank line :smiley:

PR #318, opened 2018-06-07
WIP: Resources

bogdanromanx:


src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Resources.scala
```diff
@@ -136,11 +138,57 @@ object Resources {
...
+        .values
+        .flatMap { vs =>
+          val set = vs.toSet
+          Some(set.contains(nxv.Schema) || set.contains(owl.Ontology))
+        }
+        .getOrElse(false)
+
+    def importsValues(id: AbsoluteIri, g: Graph): Set[Ref] =
+      if (canImport(id, g))
+        g.objects(_ == IriNode(id), _ == owl.imports).unorderedFoldMap {
```
umbreak: If you import `import ch.epfl.bluebrain.nexus.rdf.Graph._` you can directly do:

```scala
g.objects(IriNode(id), owl.imports)
```

PR #323, opened 2018-06-12
Added indexing of events into Blazegraph

### LINKED ISSUE #290: Index primary store events into Blazegraph

umbreak:
Fixes #290

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -25,7 +25,7 @@ scalafmt: {
  */
 
 // Dependency versions
-val commonsVersion              = "0.7.19"
+val commonsVersion              = "0.10.11"
 val rdfVersion                  = "0.2.0"
 val sourcingVersion             = "0.10.5"
 val akkaVersion                 = "2.5.12"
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -25,6 +25,7 @@ scalafmt: {
  */
 
 // Dependency versions
+val commonsVersion              = "0.7.19"
```
bogdanromanx: 0.10.x

src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/SparqlIndexer.scala
```diff
@@ -0,0 +1,83 @@
...
+      Try(rs.next().getLiteral("o").getLong).toOption
+    }
+
+  private def indexResource(res: Resource): F[Unit] =
+    materialize(res).value.flatMap {
+      case Left(err) => F.raiseError(err)
+      case Right(r)  => client.replace(res.id, r.value.graph)
+    }
+
+  private implicit def toGraphUri(id: ResId): Uri = relativeGraph.resolve(id.value)
```
bogdanromanx: let's use the id of the resource as the id of the graph so there's no need to maintain a mapping; it also goes well with the fact that we're requiring each graph to have a "primary node"

src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/SparqlIndexer.scala
```diff
@@ -0,0 +1,83 @@
...
+import scala.util.Try
+
+/**
+  * Indexer which takes a resource event and calls SPARQL client with relevant update if required
+  *
+  * @param client SPARQL client
+  */
+class SparqlIndexer[F[_]: Resolution](client: SparqlClient[F])(implicit repo: Repo[F],
+                                                               F: MonadError[F, Throwable],
+                                                               ucl: HttpClient[F, ResultSet]) {
```
bogdanromanx: is the HTTP client used? I was thinking that the sparql client is self contained.

build.sbt
```diff
@@ -67,6 +69,7 @@ lazy val monixTail            = "io.monix"                %% "monix-tail"
 lazy val pureconfig           = "com.github.pureconfig"   %% "pureconfig"                 % pureconfigVersion
 lazy val scalaTest            = "org.scalatest"           %% "scalatest"                  % scalaTestVersion
 lazy val shapeless            = "com.chuusai"             %% "shapeless"                  % shapelessVersion
+lazy val sparqlClient         = "ch.epfl.bluebrain.nexus" %% "sparql-client"              % commonsVersion
```
bogdanromanx: let's group all nexus resources to the top

src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/SparqlIndexer.scala
umbreak: When making the call to `queryRs`, the `HttpClient[F, ResultSet]` it is used

PR #328, opened 2018-06-13
Added project and organization metadata

### LINKED ISSUE #327: Add project and organization to metadata
This will be useful for federated queries, for example

umbreak:
Fixes #327

### DIFF
--- a/src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/ProjectRef.scala
+++ b/src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/ProjectRef.scala
@@ -1,8 +1,38 @@
 package ch.epfl.bluebrain.nexus.kg.resources
 
+import ch.epfl.bluebrain.nexus.rdf.Iri.Path._
+import ch.epfl.bluebrain.nexus.rdf.Iri._
+
 /**
   * A stable project reference.
   *
-  * @param id the underlying stable identifier for a project
+  * @param id The unique identifier for a project
+  * @param organization The unique identifier for the project's organization
   */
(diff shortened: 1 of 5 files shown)


src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/ProjectRef.scala
```diff
@@ -1,8 +1,38 @@
...
+
 /**
   * A stable project reference.
   *
-  * @param id the underlying stable identifier for a project
+  * @param id the unique identifier for a project
+  * @param organization the unique identifier for the project's organization
   */
-final case class ProjectRef(id: String)
+final case class ProjectRef(id: AbsoluteIri, organization: OrganizationRef) {}
```
bogdanromanx: I'm not sure about this; we need the project reference to include the _uuid_ of the project so we can use it when persisting things into the store; the _labels_ of organizations and projects need to be flexible to change later on such that we can do renames in admin without affecting the data in kg.

I do agree that the metadata is valuable, but maybe that can be handled at the routes level rather than impacting the whole codebase.

bogdanromanx: We also need to be able to create a resource id from a simple parent _uuid_ without jumping in between service calls.
I would recommend keeping the `ProjectRef` as is and just use the _uuid_s of projects and organizations; if expanding a project _uuid_ into its `AbsoluteIri` and the parent organization `AbsoluteIri` is necessary we can do it though the `P` type param of `ResourceF[P, S, A]`.

PR #332, opened 2018-06-13
Added aliasOrCurie path directive

umbreak:


src/main/scala/ch/epfl/bluebrain/nexus/kg/directives/PathDirectives.scala
```diff
@@ -0,0 +1,32 @@
...
+        case _                         => (s, "")
+      }
+      prefixMappings.get(prefix).flatMap(p => Iri.absolute(s"${p.show}$ref").toOption)
+    }
+  }
+
+  private def prefixMappings(implicit mappings: List[LoosePrefixMapping]): Map[String, AbsoluteIri] =
+    mappings.foldLeft(Map.empty[String, AbsoluteIri]) {
+      case (acc, LoosePrefixMapping(prefix, ns)) =>
+        acc + (prefix.value -> Iri.absolute(ns.value).toOption.get)
```
bogdanromanx: Option.get can throw; I recommend dropping the values that result in a None...

umbreak: It should never through because `AliasOrNamespace` is a type controlled by `refined` which can be safely converted to a `Uri`

PR #336, opened 2018-06-15
Moved additionalTypes logic to Absolute Iri analysis

umbreak:


src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/syntax.scala
```diff
@@ -7,17 +7,35 @@ import cats.Functor
...
+
     def resolveOr(f: Ref => Rejection)(implicit F: Functor[F]): EitherT[F, Rejection, Resource] =
       EitherT.fromOptionF(resolve, f(ref))
+
     def resolveAll: F[List[Resource]] = R.resolveAll(ref)
+
+    /**
+      * @return the types inferred by the [[Ref]]
+      */
+    def types: Set[AbsoluteIri] =
```
bogdanromanx: I think this should be a function in `Resources` because it doesn't really have anything to do with the `Ref` type.

bogdanromanx: We should then consider renaming it to `def inferredTypes(schema: Ref): Set[AbsoluteIri]`.

umbreak:
Yes I agree. I'm not sure if it is a requirement the ontology one. We can discuss with @MFSY . I'll keep it for now

PR #338, opened 2018-06-19
Added attachments surface API

### LINKED ISSUE #329: Management API attach binaries

umbreak:
Fixes #329

### DIFF
--- a/src/main/scala/ch/epfl/bluebrain/nexus/kg/config/AttachmentConfig.scala
+++ b/src/main/scala/ch/epfl/bluebrain/nexus/kg/config/AttachmentConfig.scala
@@ -2,5 +2,11 @@ package ch.epfl.bluebrain.nexus.kg.config
 
 import ch.epfl.bluebrain.nexus.rdf.Iri.AbsoluteIri
 
-//TODO: This should be placed inside the AppConfig case class where the app.conf file gets mapped
+/**
+  * Instance attachments specific settings.
+  * This should be placed inside the AppConfig case class where the app.conf file gets mapped
+  *
+  * @param volume          the base Iri where the attachments are stored
+  * @param digestAlgorithm the algorithm used to compute the digest of the binary
+  */
 final case class AttachmentConfig(volume: AbsoluteIri, digestAlgorithm: String)
--- a/src/main/scala/ch/epfl/bluebrain/nexus/kg/config/AttachmentConfig.scala
+++ b/src/main/scala/ch/epfl/bluebrain/nexus/kg/config/AttachmentConfig.scala
@@ -1,0 +1,6 @@
+package ch.epfl.bluebrain.nexus.kg.config
+
+import ch.epfl.bluebrain.nexus.rdf.Iri.AbsoluteIri
+
+//TODO: This should be placed inside the AppConfig case class where the app.conf file gets mapped
+final case class AttachmentConfig(volume: AbsoluteIri, digestAlgorithm: String)
--- a/src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Event.scala
+++ b/src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Event.scala
@@ -108,4 +108,32 @@ object Event {
       identity: Identity
   ) extends Event
 
+  /**
+    * A witness that a resource's attachment has been added.
+    *
+    * @param id       the resource identifier
+    * @param rev      the revision that this event generated
+    * @param value    the metadata of the attachment
+    * @param instant  the instant when this event was recorded
+    * @param identity the identity which generated this event
+    */
+  final case class Attached(id: Id[ProjectRef],
+                            rev: Long,
+                            value: BinaryAttributes,
+                            instant: Instant,
+                            identity: Identity)
+      extends Event
+
+  /**
+    * A witness that a resource's attachment has been removed.
+    *
+    * @param id       the resource identifier
+    * @param rev      the revision that this event generated
+    * @param filename the filename of the attachment removed
+    * @param instant  the instant when this event was recorded
+    * @param identity the identity which generated this event
+    */
+  final case class Unattached(id: Id[ProjectRef], rev: Long, filename: String, instant: Instant, identity: Identity)
+      extends Event
+
 }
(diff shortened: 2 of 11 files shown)


src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Repo.scala
```diff
@@ -79,6 +81,89 @@ class Repo[F[_]: Monad](agg: Agg[F], clock: Clock, toIdentifier: ResId => String
...
+    * @param attach the attachment description metadata
+    * @param source the source of the attachment
+    * @tparam In the storage input type
+    * @return either a rejection or the new resource representation in the F context
+    */
+  def attach[In](id: ResId, rev: Long, attach: BinaryDescription, source: In)(
+      implicit identity: Identity,
+      store: AttachmentStore[F, In, _]): EitherT[F, Rejection, Resource] =
+    get(id).toRight(NotFound(id.ref)).flatMap { res =>
+      if (res.deprecated) EitherT.leftT(IsDeprecated(id.ref))
```
bogdanromanx: couldn't we use `agg.checkEval` here?

src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Rejection.scala
```diff
@@ -101,4 +110,11 @@ object Rejection {
     * @param message a human readable description of the cause
     */
   final case class DownstreamServiceError(override val message: String) extends Rejection(message)
+
+  /**
+    * Signals an unexpected rejection
+    *
+    * @param message a human readable description of the cause
+    */
+  final case class Unexpected(override val message: String) extends Rejection(message)
```
bogdanromanx: maybe we should request a `F[_] : MonadError` and raise the unexpected failures with `F.raiseError` instead. It doesn't seem to be a user generated issue.

Thoughts?

src/main/scala/ch/epfl/bluebrain/nexus/kg/config/AttachmentConfig.scala
```diff
@@ -0,0 +1,6 @@
+package ch.epfl.bluebrain.nexus.kg.config
+
+import ch.epfl.bluebrain.nexus.rdf.Iri.AbsoluteIri
+
+//TODO: This should be placed inside the AppConfig case class where the app.conf file gets mapped
+final case class AttachmentConfig(volume: AbsoluteIri, digestAlgorithm: String)
```
bogdanromanx: docs?

src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Event.scala
```diff
@@ -108,4 +109,32 @@ object Event {
...
+  /**
+    * A witness that a resource's attachment has been added.
+    *
+    * @param id       the resource identifier
+    * @param rev      the revision that this event generated
+    * @param value    the metadata of the attachment
+    * @param instant  the instant when this event was recorded
+    * @param identity the identity which generated this event
+    */
+  final case class Attached(id: Id[ProjectRef],
```
bogdanromanx: Could we rename the event types to `AttachmentAdded` and `AttachmentRemoved` to match the `Tag*` naming?

bogdanromanx: Same for commands...

src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Repo.scala
umbreak: Yes, but in practice it becomes more ugly because we have to create a dummy command and event for that. I think this is a neater way.

src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Rejection.scala
umbreak: I didn't want to mix `MonadError` in some places with `EitherT` or `OptionT` in others. I think having a common one makes things cleaner. There is nowhere on the definition of a `Rejection` an assumption of user input errors only (DownsrteamRejection as another example of non user input error).

bogdanromanx: I think longer term would be better to discriminate between system failures and user rejections.
We still have the case where we don't produce an EitherT value; for example in sourcing we produce a failed future equivalent to `F.raiseError`:

https://github.com/BlueBrain/nexus-sourcing/blob/master/modules/akka/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/akka/ShardingAggregate.scala#L46

src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Repo.scala
bogdanromanx: We would have to create a dummy instance of a command, but no events:

https://github.com/BlueBrain/nexus-sourcing/blob/b93baefd83d5bf053fc83cdd16d32055a927ac55/modules/core/src/main/scala/ch/epfl/bluebrain/nexus/sourcing/Aggregate.scala#L39

But I see your point, it's fine to keep it like this.

src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/attachment/AttachmentStore.scala
```diff
@@ -0,0 +1,170 @@
...
+      *
+      * @param loc    the location of the attachment
+      * @param source the source
+      * @param meta   the source metadata
+      */
+    def toSink(loc: Location, source: In, meta: BinaryDescription): EitherT[F, Rejection, StoredSummary]
+  }
+
+  object Stream {
+    def akka(config: AttachmentConfig)(implicit as: ActorSystem): Stream[Future, AkkaIn, AkkaOut] =
```
bogdanromanx: We should also provide a Stream[Task, AkkaIn, AkkaOut] that wraps the akka one, something like:
```
def task(config: AttachmentConfig)(implicit as: ActorSystem): Stream[Task, AkkaIn, AkkaOut] =
  new Stream[Task, AkkaIn, AkkaOut] {
    val underlying = akka(config)
    def toSource(uri: AbsoluteIri): Either[Rejection, AkkaOut] =
      underlying.toSource(uri)
    def toSink(...): EitherT[Task, Rejection, StoredSummary] = {
      EitherT(Task.deferFuture(underlying.toSink(...).value))
    }
  }
```
such that we can use it directly in the routes using the same `F`.

bogdanromanx: We can do this separately, though.

PR #347, opened 2018-06-26
 Do not validate against SHACL schema for now and added types to deprecate events

umbreak:


src/test/resources/application.conf
```diff
@@ -0,0 +1,9 @@
+akka {
+  log-dead-letters = off
+  stdout-loglevel = "OFF"
+  log-dead-letters-during-shutdown = off
+  loggers = ["akka.event.slf4j.Slf4jLogger"]
+  logging-filter = "akka.event.slf4j.Slf4jLoggingFilter"
+  loglevel = INFO
+  loglevel = ${?LOG_LEVEL}
```
hygt: Not sure we need the environment variable in the test config.

PR #381, opened 2018-07-05
Index everything

hygt:
* Accounts
* Projects
* Views
* Resolvers


src/main/scala/ch/epfl/bluebrain/nexus/kg/async/ProjectViewCoordinator.scala
```diff
@@ -93,14 +91,16 @@ class ProjectViewCoordinator(projects: Projects[Future], actorCtor: View => Prop
...
+        ref ! Stop
+        context.stop(ref)
       }
 
       // construct actors for the new view updates
       if (added.nonEmpty) log.debug(s"Creating view coordinators for $added")
       val newActorsMapping = withNames
         .filter { case (name, _) => added.contains(name) }
-        .mapValues(v => context.actorOf(actorCtor(v)))
+        .mapValues(v => selector(v))
```
umbreak:
```scala
.mapValues(selector)
```

PR #391, opened 2018-07-11
Implement listings.

wwajerowicz:


src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/ElasticDecoders.scala
```diff
@@ -0,0 +1,63 @@
...
+import ch.epfl.bluebrain.nexus.kg.config.AppConfig.HttpConfig
+import ch.epfl.bluebrain.nexus.rdf.Iri
+import ch.epfl.bluebrain.nexus.rdf.Iri.AbsoluteIri
+import ch.epfl.bluebrain.nexus.rdf.akka.all._
+import io.circe.Decoder
+
+object ElasticDecoders {
+
+  /**
+    * Circe decoder which which reconstructs resource representation ID from ElasticSearch response
```
hygt: which which :smile:

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/ResourceRoutes.scala
```diff
@@ -6,15 +6,20 @@ import akka.http.javadsl.server.Rejections.validationRejection
 import akka.http.scaladsl.model.ContentTypes.`application/octet-stream`
 import akka.http.scaladsl.model.headers.RawHeader
 import akka.http.scaladsl.model.{ContentType, HttpEntity, StatusCodes}
-import akka.http.scaladsl.server.Directives._
+import akka.http.scaladsl.server.Directives.{parameter, _}
```
hygt: redundant import

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/ResourceRoutes.scala
```diff
@@ -289,7 +312,11 @@ object ResourceRoutes {
                     iamClient: IamClient[Task],
                     indexers: IndexerClients[Task],
                     store: AttachmentStore[Task, AkkaIn, AkkaOut],
-                    config: AppConfig): ResourceRoutes = new ResourceRoutes()
+                    config: AppConfig,
+                    elasticClient: ElasticClient[Task],
```
umbreak: the elasticClient already exists inside indexers. Why passing it 2 times?

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/ResourceRoutes.scala
```diff
@@ -48,7 +54,11 @@ class ResourceRoutes(implicit repo: Repo[Task],
                      iamClient: IamClient[Task],
                      indexers: IndexerClients[Task],
                      store: AttachmentStore[Task, AkkaIn, AkkaOut],
-                     config: AppConfig) {
+                     config: AppConfig,
+                     elasticClient: ElasticClient[Task],
```
umbreak: the elasticClient already exists inside indexers. Why passing it 2 times?

src/main/scala/ch/epfl/bluebrain/nexus/kg/search/QueryResultEncoder.scala
```diff
@@ -0,0 +1,48 @@
...
+        }: _*
+      )
+    )
+  }
+
+  /**
+    * Encoder for unscored query results
+    */
+  implicit val unscoredEncoder: Encoder[UnscoredQueryResults[AbsoluteIri]] = Encoder.instance { results =>
+    Json.obj(
```
umbreak: you can just create the Json without the `@context` and then use `.addContext(Contexts.searchCtxUri)`

umbreak: Here you built a Json, while the preferred way is to build a graph and output it in Json format (as we did when outputing `ResourceF` or `ResourceV`. We ensure like this json-LD validity

src/main/scala/ch/epfl/bluebrain/nexus/kg/search/QueryResultEncoder.scala
```diff
@@ -0,0 +1,48 @@
...
+import ch.epfl.bluebrain.nexus.rdf.Iri.AbsoluteIri
+import io.circe.{Encoder, Json}
+
+object QueryResultEncoder {
+
+  /**
+    * Encoder for scored query results
+    */
+  implicit val scoredEncoder: Encoder[ScoredQueryResults[AbsoluteIri]] = Encoder.instance { results =>
+    Json.obj(
```
umbreak: Same here. use `.addContext(Contexts.searchCtxUri)` and use a graph instead of building a pure json

src/test/resources/resources/elastic-resource.json
```diff
@@ -0,0 +1,6 @@
+{
+  "@id": "http://resources.nexus.com/test-resource/306e8f68-8419-11e8-adc0-fa7ae01bbebc",
+  "project": "http://admin.nexus.example.com/projects/bbp/testproject",
```
umbreak: Where this project is coming from? so far only [this metadata](https://github.com/wwajerowicz/nexus-kg/blob/685e3cfa2d068abb7d4346008c1dca595e225d84/src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/ResourceF.scala#L74) is stored on a ElasticSearch document

wwajerowicz: I missed that, I need to add it to the indexer.

src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/ElasticDecoders.scala
```diff
@@ -0,0 +1,42 @@
...
+        id <- json("@id").flatMap(_.asString).map(Iri.absolute).getOrElse(Left("Field: '@id' not found"))
+        schema <- json("constrainedBy")
+          .flatMap(_.asString)
+          .map(Iri.absolute)
+          .getOrElse(Left("Field: 'constrainedBy' not found"))
+        reprId = prefix + aliasOrCurieFor(schema, project) + aliasOrCurieFor(id, project)
+      } yield reprId
+    }
+
+  private def aliasOrCurieFor(iri: AbsoluteIri, project: Project): String = {
```
umbreak: You can simplify it with something like this:
```scala
project.prefixMappings.collectFirst { 
      case (prefix, ns) if iri.show.startsWith(ns.show) =>
         s"$prefix:${iri.show.stripPrefix(matchedIri.show)}"
       case (prefix, ns) if iri == ns => 
         prefix
}.getOrElse(iri.show)
```

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/ResourceRoutes.scala
```diff
@@ -129,6 +140,21 @@ class ResourceRoutes(implicit repo: Repo[Task],
         }
     }
 
+  private def listings(implicit token: Option[AuthToken]): Route =
+    (pathPrefix("resources") & project) { implicit proj =>
+      implicit val esClient = indexers.elastic
+      implicit val decoder = ElasticDecoder.apply(ElasticDecoders.resourceIdDecoder(url"${config.http.publicUri.copy(
```
umbreak: `import ch.epfl.bluebrain.nexus.service.http.UriOps._`
```scala
config.http.publicUri.append(/ "resources" / proj.label.organizationReference / proj.label.projectLabel)
```

src/test/resources/resources/elastic-resource.json
umbreak: I would say the project uuid is needed as a metadata, but not the project URI (because it might change, since the labels can be updated).

In any case, for your logic, you should not need the project URI anywhere, since in the routes you already have the project label and the account label

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/ResourceRoutes.scala
```diff
@@ -42,10 +48,12 @@ import scala.concurrent.Future
 import scala.util.Try
 
 class ResourceRoutes(implicit repo: Repo[Task],
-                     projects: Projects[Task],
                      indexers: Clients[Task],
                      store: AttachmentStore[Task, AkkaIn, AkkaOut],
-                     config: AppConfig) {
+                     config: AppConfig,
+                     httpClient: UntypedHttpClient[Task],
```
umbreak: the `httpClient` and `mt` I guess are needed in order to generate the `HttpClient`.
I would put them inside `Clients` case class in the implicit part

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/ResourceRoutes.scala
```diff
@@ -282,10 +305,12 @@ class ResourceRoutes(implicit repo: Repo[Task],
 
 object ResourceRoutes {
   final def apply()(implicit repo: Repo[Task],
-                    projects: Projects[Task],
                     indexers: Clients[Task],
                     store: AttachmentStore[Task, AkkaIn, AkkaOut],
-                    config: AppConfig): ResourceRoutes = new ResourceRoutes()
+                    config: AppConfig,
+                    httpClient: UntypedHttpClient[Task],
```
umbreak: sam here for `httpClient` and `mt`

src/main/scala/ch/epfl/bluebrain/nexus/kg/search/QueryResultEncoder.scala
```diff
@@ -13,39 +20,74 @@ object QueryResultEncoder {
...
 
+  private def addResults[A <: QueryResult[AbsoluteIri]](
+      graph: Graph,
+      mainNode: IriOrBNode,
+      results: QueryResults[AbsoluteIri],
+      triples: (QueryResult[AbsoluteIri], BNode) => Set[Graph.Triple]): Json = {
+    val graphWithList = results.results match {
+      case Nil => graph.add(mainNode, nxv.results, rdf.nil)
+      case first :: rest =>
+        val firstNode     = BNode()
```
umbreak: this case block is the same as the case block inside the fold left. Why don't you provide an inner function and refactor it together?

src/main/scala/ch/epfl/bluebrain/nexus/kg/search/QueryResultEncoder.scala
```diff
@@ -13,39 +20,74 @@ object QueryResultEncoder {
...
+
+            val graphWithNext = g ++ Graph(triples(result, resultNode))
+              .add(listNode, rdf.rest, nextListNode)
+              .add(nextListNode, rdf.first, resultNode)
+
+            (graphWithNext, nextListNode)
+        }
+        graphWithoutListEnd.add(lastElement, rdf.rest, rdf.nil)
+    }
+    graphWithList
```
umbreak:
```scala
graph.asJson(searchCtx).getOrElse(graph.asJson()).removeKeys("@context", "@id").addContext(searchCtxUri)
```

src/main/scala/ch/epfl/bluebrain/nexus/kg/search/QueryResultEncoder.scala
wwajerowicz: It's not really the same, here it links the main node with the first list node and inside `foldLeft` it links two list nodes together.

src/main/scala/ch/epfl/bluebrain/nexus/kg/search/QueryResultEncoder.scala
```diff
@@ -83,11 +83,17 @@ object QueryResultEncoder {
         }
         graphWithoutListEnd.add(lastElement, rdf.rest, rdf.nil)
     }
+    import ch.epfl.bluebrain.nexus.rdf.syntax.nexus._
     graphWithList
       .asJson(Contexts.searchCtx)
```
umbreak: Whenever you know the `@id` (in your case this localhost trick), you should use it instead. So if you use:
```scala
graphWithList.asJson(searchCtx, url"http://localhost")
```
You can remove `import ch.epfl.bluebrain.nexus.rdf.syntax.nexus._` and put the other import on top without this private method

PR #392, opened 2018-07-11
Added cache resolution from labels

umbreak:


src/main/scala/ch/epfl/bluebrain/nexus/kg/async/Projects.scala
```diff
@@ -195,17 +235,30 @@ object Projects {
...
+      def updateAccount() = {
+        val empty  = LWWRegister(RevisionedValue[Option[Account]](0L, None))
+        val value  = RevisionedValue[Option[Account]](ac.rev, Some(ac))
+        val update = Update(accountKey(ref), empty, WriteMajority(tm.duration))(_.withValue(value))
+        (replicator ? update).flatMap(handleBooleanUpdate("Timed out while waiting for add project quorum response"))
+      }
+
+      def updateAccountLabel() = {
+        val empty                                  = LWWRegister(RevisionedValue[Option[String]](0L, None))
+        val value: RevisionedValue[Option[String]] = RevisionedValue(ac.rev, Some(ac.label))
```
hygt: Any reason for the explicit type?

PR #393, opened 2018-07-12
Used cache to retrieve project

### LINKED ISSUE #362: Use cached project instead of calling ADMIN in routes
Every time a client hits an endpoint, the routes are calling `admin` service to fetch the project (the `prefixMappings` are used to resolve curies or aliases).

This can be prevented, since we have the project and account information on the `cache`.
However the key of the cache is the `project_uuid` and not the `project_label` while in the routes what we have is the `project_label`.

A solution would be to add another cache, or a simple custerSingleton which is just a `Map[String, String]` where the keys are the `project_label` and the values are the `project_uuid`.

umbreak:
Fixes #362


src/main/scala/ch/epfl/bluebrain/nexus/kg/directives/ProjectDirectives.scala
```diff
@@ -16,14 +17,22 @@ object ProjectDirectives {
   /**
     * Fetches project configuration from nexus admin
```
hygt: Fetches project configuration from the cache if possible, from nexus admin otherwise.

PR #395, opened 2018-07-13
Create ES mapping when starting indexing.

### LINKED ISSUE #385: Define static prefix mappings for metadata fields in default ES view
The [static mapping](https://github.com/BlueBrain/nexus-iam/blob/master/modules/elastic/src/main/resources/elastic-mappings.json) for the key values of the metadata has to be created and the index itself has to be created with that mapping on [ElasticIndexer](https://github.com/BlueBrain/nexus-kg/blob/master/src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/ElasticIndexer.scala#L95)

wwajerowicz:
Fixes #385

### DIFF
--- a/src/main/resources/elastic/mapping.json
+++ b/src/main/resources/elastic/mapping.json
@@ -1,0 +1,19 @@
+{
+  "mappings": {
+    "{{docType}}": {
+      "properties": {
+        "@type": {"type": "keyword"},
+        "@id": {"type": "keyword"},
+        "_rev": {"type": "keyword"},
+        "_deprecated": {"type": "boolean"},
+        "_createdAt": {"type": "date"},
+        "_updatedAt": {"type": "date"},
+        "_createdBy": {"type": "keyword"},
+        "_updatedBy": {"type": "keyword"},
+        "_constrainedBy": {"type": "keyword"},
+        "_source": {"type": "text"}
+      },
+      "dynamic": false
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


src/main/resources/elastic/mapping.json
```diff
@@ -0,0 +1,19 @@
+{
+  "mappings": {
+    "{{docType}}": {
+      "properties": {
+        "@type": {"type": "keyword"},
+        "@id": {"type": "keyword"},
+        "_rev": {"type": "keyword"},
```
umbreak:
```json
"_rev": {"type": "long"}
```

src/main/resources/elastic/mapping.json
```diff
@@ -0,0 +1,19 @@
...
+        "_deprecated": {"type": "boolean"},
+        "_createdAt": {"type": "date"},
+        "_updatedAt": {"type": "date"},
+        "_createdBy": {"type": "keyword"},
+        "_updatedBy": {"type": "keyword"},
+        "_constrainedBy": {"type": "keyword"},
+        "_source": {"type": "text"}
```
umbreak: What about the `_original` field ?

wwajerowicz: I renamed `original` to `_source` as I thought that would be more meaningful.

umbreak: But it isn't renamed right...it is still [here](https://github.com/wwajerowicz/nexus-kg/blob/a366ebb21a86dc131bf3ec323d8449b79b2e60cb/src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/ElasticIndexer.scala#L71)

wwajerowicz: Yes, I thought that I renamed it, but I only added `_` 😄

PR #402, opened 2018-07-16
Add in account resolution

hygt:


src/main/scala/ch/epfl/bluebrain/nexus/kg/resolve/InAccountResolution.scala
```diff
@@ -0,0 +1,52 @@
...
+import cats.syntax.traverse._
+import ch.epfl.bluebrain.nexus.kg.async.DistributedCache
+import ch.epfl.bluebrain.nexus.kg.resources._
+
+/**
+  * Resolution implementation that looks into every project belonging to a given account, in no particular order.
+  *
+  * TODO: ACLs
+  */
+class InAccountResolution[F[_]](accountRef: AccountRef, resources: Resources[F], cache: DistributedCache[F])(
```
umbreak: I think we should be consistent with the way we build those resolvers. `CrossProjectResolver` does not depend on `Projects` (or `DistributedCache `) as you called it. It just depends on a `List[Resolvers]` what gets passed to the constructor. In this way, this class does not have to know about the cache at all.

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/DistributedCache.scala
```diff
@@ -21,9 +21,10 @@ import monix.eval.Task
 import scala.concurrent.{ExecutionContext, Future}
 
 /**
-  * Project specific operations.
+  * Contract for the distributed cache that keeps in-memory metadata and content
+  * of the resources indexed by the service.
   */
-trait Projects[F[_]] {
+trait DistributedCache[F[_]] {
```
umbreak: It is not a big deal, but I considered this trait as `bundle operations for Projects`, as an analogy to what `Resources` trait is for any generic resource. 

Calling it `DistributedData` is only exposing information about the technique we use to deal with the data, but not about what type of data/operation this trait takes care of.

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/DistributedCache.scala
```diff
@@ -273,27 +292,57 @@ object Projects {
...
+                handleBooleanUpdate("Timed out while waiting for add project quorum response"))
+            case _ => Future.successful(false)
+          }
+        }
+      }
+    }
+
+    private def addProjectToAccount(ref: ProjectRef, accountRef: AccountRef, instant: Instant): Future[Boolean] = {
+      projects(accountRef).flatMap { projects =>
+        if (projects.contains(ref)) Future.successful(true)
```
umbreak: When it already exists, do we signal it as true ? I think [in some places](https://github.com/hygt/nexus-kg/blob/9a03abf7d8bb14f9395758390b714abd3d0bae95/src/main/scala/ch/epfl/bluebrain/nexus/kg/async/DistributedCache.scala#L412) we had the opposite logic (return false if it already exists). Seems that we only return true when actually something gets added/deleted/updated and not when the thing is already existing

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/DistributedCache.scala
bogdanromanx: If it contains operations related to projects only we should preserve its name, otherwise if it includes things outside the realm of projects then Cache is probably a better name.

src/main/scala/ch/epfl/bluebrain/nexus/kg/resolve/InAccountResolution.scala
bogdanromanx: It depends on how this class is to be used, and quite important. If we want to construct a single instance for an account and pass it around it's fine then, otherwise I agree with the above. But we should be consistent in the implementations.

src/main/scala/ch/epfl/bluebrain/nexus/kg/resolve/ProjectResolution.scala
```diff
@@ -19,7 +19,8 @@ import scala.concurrent.{ExecutionContext, Future}
   * @param staticResolution the static resolutions
   * @tparam F the monadic effect type
   */
-abstract class ProjectResolution[F[_]: Monad](cache: DistributedCache[F], staticResolution: Resolution[F]) {
+abstract class ProjectResolution[F[_]](cache: DistributedCache[F], staticResolution: Resolution[F])(
```
umbreak: you can keep
```scala
import cats.syntax.applicative._
abstract class ProjectResolution[F[_]: Monad](cache: DistributedCache[F], staticResolution: Resolution[F]) {
```

src/main/scala/ch/epfl/bluebrain/nexus/kg/resolve/ProjectResolution.scala
```diff
@@ -34,11 +35,15 @@ abstract class ProjectResolution[F[_]: Monad](cache: DistributedCache[F], static
...
-          case _: CrossProjectResolver => CrossProjectResolution[F](resources, res)
+        val result = res.filterNot(_.deprecated).toList.sortBy(_.priority).map {
+          case _: StaticResolver    => staticResolution
+          case r: InProjectResolver => InProjectResolution[F](r.ref, resources)
+          case r: InAccountResolver =>
+            val projects = cache.projects(r.accountRef).map(_.map(_ -> r.resourceTypes).toList)
+            MultiProjectResolution(resources, projects)
+          case r: CrossProjectResolver =>
+            val projects = r.projects.map(_ -> r.resourceTypes).toList
+            MultiProjectResolution(resources, F.pure(projects))
```
umbreak: `projects.pure`

src/main/scala/ch/epfl/bluebrain/nexus/kg/resolve/ProjectResolution.scala
```diff
@@ -34,11 +35,15 @@ abstract class ProjectResolution[F[_]: Monad](cache: DistributedCache[F], static
...
-        val result = res.map {
-          case _: StaticResolver       => staticResolution
-          case r: InProjectResolver    => InProjectResolution[F](r.ref, resources)
-          case r: InAccountResolver    => InAccountResolution[F](r.accountRef, resources, cache)
-          case _: CrossProjectResolver => CrossProjectResolution[F](resources, res)
+        val result = res.filterNot(_.deprecated).toList.sortBy(_.priority).map {
+          case _: StaticResolver    => staticResolution
+          case r: InProjectResolver => InProjectResolution[F](r.ref, resources)
+          case r: InAccountResolver =>
+            val projects = cache.projects(r.accountRef).map(_.map(_ -> r.resourceTypes).toList)
```
umbreak: `toList.distinct` ? Since there is room for duplication and we should avoid calling twice the resolution for the same project and resource type

PR #417, opened 2018-07-23
Added InProject and InAccount resolvers

umbreak:


src/main/scala/ch/epfl/bluebrain/nexus/kg/search/QueryResultEncoder.scala
```diff
@@ -45,6 +45,7 @@ object QueryResultEncoder {
       g.asJson(searchCtx deepMerge extraCtx, Some(mainNode))
         .getOrElse(g.asJson)
         .removeKeys("@context", "@id")
+        .removeKeys("@context", "@id")
```
bogdanromanx: ?

src/test/scala/ch/epfl/bluebrain/nexus/kg/TestHelper.scala
```diff
@@ -40,4 +43,26 @@ trait TestHelper {
 
   def uuid = UUID.randomUUID().toString
 
+  implicit val jsonEq: Equality[Json] = new Equality[Json] {
```
bogdanromanx: this is dangerous as order of elements in an array it's important; it may make sense for json-ld if the container type is a `@set`

I would recommend implementing a word which allows choosing when the order doesn't matter

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/DistributedCache.scala
```diff
@@ -299,25 +325,26 @@ object DistributedCache {
...
-            case _ => Future.successful(false)
-          }
-        }
+    private def updateProject(ref: ProjectRef,
+                              accountRef: AccountRef,
+                              proj: Project,
+                              instant: Instant,
+                              updateRev: Boolean): Future[Boolean] = {
+      val result = for {
+        _ <- updateProject(ref, proj).withFilter(identity)
```
bogdanromanx: `r <- updateProject(ref, proj) if r` is probably more readable...

src/main/scala/ch/epfl/bluebrain/nexus/kg/directives/LabeledProject.scala
```diff
@@ -1,13 +1,18 @@
...
-  *                containing the project label and organization label
-  * @param project the project
+  * @param label      the project reference information,
+  *                   containing the project label and organization label
+  * @param project    the project
+  * @param accountRef the account reference
   */
-final case class LabeledProject(label: ProjectLabel, project: Project)
+final case class LabeledProject(label: ProjectLabel, project: Project, accountRef: AccountRef) {
+  def ref: ProjectRef   = ProjectRef(project.uuid)
```
bogdanromanx: docs?

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/DistributedCache.scala
umbreak: Then I have to assign variables for each of the for loop elements? I don't find it much easier

hygt: I agree with Bogdan, `withFilter` in a for comprehension is weird.

PR #418, opened 2018-07-24
Fix ES indexing.

wwajerowicz:


src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/ElasticIndexer.scala
```diff
@@ -65,14 +68,21 @@ class ElasticIndexer[F[_]](client: ElasticClient[F], index: String, resources: R
...
 
     val payload = graph.asJson(resourceCtx, primaryNode).getOrElse(graph.asJson)
-    val merged  = payload deepMerge Json.obj("_source" -> Json.fromString(res.value.noSpaces))
-    client.update(index, config.docType, res.id.elasticId, merged)
+    val merged  = payload deepMerge Json.obj("_original_source" -> Json.fromString(res.value.noSpaces))
+    val esPayload = Json.obj(
+      "doc"           -> merged,
+      "doc_as_upsert" -> Json.fromBoolean(true)
+    )
+    client.update(index, config.docType, res.id.elasticId, esPayload)
```
umbreak: Am I missing something or we are doing an update, while we should be doing a full replace?
From the docs:
> The update API also support passing a partial document, which will be merged into the existing document (simple recursive merge, inner merging of objects, replacing core "keys/values" and arrays). To fully replace the existing document, the index API should be used instead. The following partial update adds a new field to the existing document

wwajerowicz: You're right, we should be doing what this [doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/_modifying_your_data.html) says, but instead we're doing an [_update](https://www.elastic.co/guide/en/elasticsearch/reference/current/_updating_documents.html). I'll fix that.

umbreak: Yes, it is a regular [create call](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html)

src/main/scala/ch/epfl/bluebrain/nexus/kg/marshallers/instances/package.scala
```diff
@@ -40,8 +40,20 @@ package object instances extends FailFastCirceSupport {
   final implicit def httpEntity[A](implicit encoder: Encoder[A],
                                    printer: Printer = Printer.noSpaces.copy(dropNullValues = true),
                                    keys: OrderedKeys = OrderedKeys(
-                                     List("@context", "@id", "@type", "code", "message", "details", "")))
-    : ToEntityMarshaller[A] =
+                                     List("@context",
```
umbreak: I guess we could add the default behaviour of the orderedKeys directly in `AppConfig` as implicit val, since it can be reused anywhere we need it

PR #423, opened 2018-07-26
Prevent from having several InAccount and InProject resolvers

umbreak:


bogdanromanx:
I'm unsure if we want this restriction in the system as there may be (some) cases where we may want to have multiple resolvers of the same type with different priorities targeting different resource types.

Parked for now.

PR #426, opened 2018-07-27
Set projects in CrossProjectResolver as account/project

### LINKED ISSUE #419: Cross project API call without having to specify the project UUID
A cross project resolver, from the API consumer side, looks like this now:
```json
{
  "@type": [ "Resolver", "CrossProject" ],
  "resourceTypes": [ "nxv:Schema","nxv:Resource"],
  "projects": [
    "729fa67d-f896-4802-a7b5-7da85555097b",
    "93e9dc8e-e118-4f73-ab1c-2cb93151048f"
  ],
  "priority": 12,
  "identities": [
    {
                "@type": "GroupRef",
                "realm": "bbp-test",
                "group": "bbp-ou-nexus"
    }
  ]
}
```

The `projects` field contains an array of UUIDs. We have to change it, so that the client can write the projects block like this:

```json
  "projects": [
    "account1/project1",
    "account2/project3"
  ],
```

In the cache we already have the mechanism in place to translate segments into UUIDs

umbreak:
Fixes #419

### DIFF
--- a/src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/AdditionalValidation.scala
+++ b/src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/AdditionalValidation.scala
@@ -1,12 +1,23 @@
 package ch.epfl.bluebrain.nexus.kg.resources
 
-import cats.Applicative
 import cats.data.EitherT
+import cats.{Applicative, Monad}
+import ch.epfl.bluebrain.nexus.commons.http.syntax.circe._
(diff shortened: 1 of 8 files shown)


src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/AdditionalValidation.scala
```diff
@@ -29,7 +43,7 @@ object AdditionalValidation {
     * @return a new validation that always returns Right(()) on the provided effect type
```
hygt: Update scaladoc

Issue #367, opened 2018-07-03
Unit tests


umbreak:
Increase the coverage of the project at least to 80%, covering tests for:
- Surface API
- HTTP routes
- Indexing
- Listing

umbreak:
We have currently unit tests at > 85% coverage. I close it

PR #433, opened 2018-08-03
Increased coverage and added bug fixing

umbreak:
Partially addresses #367


src/main/scala/ch/epfl/bluebrain/nexus/kg/async/RevisionedValue.scala
```diff
@@ -1,17 +1,9 @@
...
   * A value with an attached revision corresponding to the value.
   *
   * @param rev   the value revision
   * @param value the value
   */
 final case class RevisionedValue[A](rev: Long, value: A) extends RegisteredValue[A]
-
-object RevisionedValue {
-
-  private[async] def revisionedValueClock[A]: Clock[RevisionedValue[A]] =
```
bogdanromanx: Why the removal?

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/RevisionedValue.scala
```diff
@@ -1,17 +1,9 @@
...
 
-import akka.cluster.ddata.LWWRegister.Clock
-
 /**
   * A value with an attached revision corresponding to the value.
   *
   * @param rev   the value revision
   * @param value the value
   */
 final case class RevisionedValue[A](rev: Long, value: A) extends RegisteredValue[A]
```
bogdanromanx: I would recommend renaming `RegisteredValue[A]` to `RegisterValue[A]`.

bogdanromanx: And possibly make it a sealed trait while moving the sub types in the same file.

src/main/scala/ch/epfl/bluebrain/nexus/kg/resolve/ResolverEncoder.scala
```diff
@@ -1,21 +1,29 @@
 package ch.epfl.bluebrain.nexus.kg.resolve
 
+import ch.epfl.bluebrain.nexus.commons.types.search.{QueryResult, QueryResults}
 import ch.epfl.bluebrain.nexus.iam.client.types.Identity
 import ch.epfl.bluebrain.nexus.iam.client.types.Identity.{AuthenticatedRef, GroupRef, UserRef}
+import ch.epfl.bluebrain.nexus.kg.config.Contexts.{resolverCtx, _}
```
bogdanromanx: fix import

src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Ref.scala
```diff
@@ -41,10 +41,13 @@ object Ref {
...
         val (q, opt) = extractTagRev(query)
-        refOf(u.copy(query = Some(q)), opt)
+        val qry      = if (q.value.isEmpty) None else Some(q)
+        refOf(u.copy(query = qry), opt)
       case u @ Urn(_, _, _, Some(query), _) =>
         val (q, opt) = extractTagRev(query)
-        refOf(u.copy(q = Some(q)), opt)
+        val qry      = if (q.value.isEmpty) None else Some(q)
+        refOf(u.copy(q = qry), opt)
+
```
bogdanromanx: empty line

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/RevisionedValue.scala
umbreak: It wasn't used. When we use it I will add it, but right now is just garbage

bogdanromanx: It is used, but not immediately obvious as creating a register takes a default clock implicitly:
```
/**
   * INTERNAL API
   */
  @InternalApi private[akka] def apply[A](node: UniqueAddress, initialValue: A, clock: Clock[A]): LWWRegister[A] =
    new LWWRegister(node, initialValue, clock(0L, initialValue))

  def apply[A](initialValue: A)(implicit node: Cluster, clock: Clock[A] = defaultClock[A]): LWWRegister[A] =
    apply(node.selfUniqueAddress, initialValue, clock)
```

Also, the only reason this type exists is to be used in a LWWRegister; if the clock is "garbage" the whole type is "garbage".

PR #437, opened 2018-08-06
Used revision as a counter for all the cache set

### LINKED ISSUE #434: Make view updates independent of timestamps
Same as done for [resolvers](https://github.com/BlueBrain/nexus-kg/pull/428)

umbreak:
Fixes #434


src/main/scala/ch/epfl/bluebrain/nexus/kg/async/DistributedCache.scala
```diff
@@ -183,61 +175,74 @@ trait DistributedCache[F[_]] {
...
-    else addView(ref, view, instant, updateRev = true)
+  def applyView(ref: ProjectRef, view: View): F[Boolean] =
+    if (view.deprecated) removeView(ref, view.id, view.rev)
+    else addView(ref, view, updateRev = true)
 }
 
 object DistributedCache {
 
-  private[async] def accountKey(ref: AccountRef): LWWRegisterKey[RevisionedValue[Option[Account]]] =
+  private[async] sealed trait RevisionedId[A] {
```
bogdanromanx: I would suggest implementing this as a type class (see https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Show.scala) and provide instances in the respective companion objects.

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/DistributedCache.scala
```diff
@@ -341,27 +344,39 @@ object DistributedCache {
...
-          val empty  = LWWRegister(TimestampedValue(0L, Set.empty[ProjectRef]))
-          val value  = TimestampedValue(instant.toEpochMilli, projects + ref)
-          val update = Update(accountProjectsKey(accountRef), empty, WriteMajority(tm.duration))(_.withValue(value))
+          val empty = LWWRegister(RevisionValue(0L, Set.empty[ProjectRef]))
+          val update = Update(accountProjectsKey(accountRef), empty, WriteMajority(tm.duration)) { currentState =>
+            val currentRevision = currentState.value.rev
+            val currentValue    = currentState.value.value
+            currentValue.find(_.id == ref.id) match {
+              case Some(r) =>
+                currentState.withValue(RevisionValue(currentRevision + 1, currentValue - r + ref))
```
bogdanromanx: Relying on a simple counter is dangerous when performing writes from multiple nodes as it won't converge in concurrent modifications. Please create a ticket for a better implementation based on ORSets or custom data type.

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/RevisionValue.scala
```diff
@@ -0,0 +1,17 @@
...
+
+import akka.cluster.ddata.LWWRegister.Clock
+
+/**
+  * A value with an attached revision corresponding to the value.
+  *
+  * @param rev   the value revision
+  * @param value the value
+  */
+final case class RevisionValue[A](rev: Long, value: A)
```
bogdanromanx: I'd recommend keeping the previous name (_RevisionedValue_) as is better conveys the data type (a value with an attached revision). It's also consistent with _RevisionedId_.

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/DistributedCache.scala
bogdanromanx: You can also move it to the package level as we can expect uses outside the scope of the cache.

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/RevisionedId.scala
```diff
@@ -0,0 +1,28 @@
+package ch.epfl.bluebrain.nexus.kg.async
+
+import ch.epfl.bluebrain.nexus.rdf.Iri.AbsoluteIri
+
+/**
+  * A type class of ''A'' which has the methods ''id'' and ''rev''
```
bogdanromanx: _A type class that provides a specific revision identifier (''id: AbsoluteIri'' and ''rev: Long'') of a resource of type ''A''._

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/RevisionedId.scala
```diff
@@ -0,0 +1,28 @@
...
+  * with the provided signature
+  */
+sealed trait RevisionedId[A] {
+  def id(a: A): AbsoluteIri
```
bogdanromanx: add scaladoc: _Returns the unique identifier of the resource 'a'_

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/RevisionedId.scala
```diff
@@ -0,0 +1,28 @@
...
+  def rev(a: A): Long
```
bogdanromanx: add scaladoc: _Returns the revision number of the identified resource 'a'_

PR #452, opened 2018-08-13
Indexing bugfixes

wwajerowicz:


src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/ResourceRoutes.scala
```diff
@@ -222,13 +222,13 @@ class ResourceRoutes(resources: Resources[Task])(implicit cache: DistributedCach
...
     (pathPrefix("resources") & project) { implicit wrapped =>
-      (get & parameter('deprecated.as[Boolean].?) & paginated & hasPermission(resourceRead)) {
+      (get & parameter('deprecated.as[Boolean].?) & paginated & hasPermission(resourceRead) & pathEndOrSingleSlash) {
         (deprecated, pagination) =>
           val results = cache.views(wrapped.ref).flatMap(v => resources.list(v, deprecated, pagination))
           trace("listResources") {
             complete(results.runAsync)
           }
-      } ~ pathPrefix(aliasOrCurie)(listings)
+      } ~ (pathPrefix(aliasOrCurie) & pathEndOrSingleSlash)(listings)
```
umbreak: pathEndOrSingleSlash shouldn't go here. It has to be handled by the `listings` method

src/main/scala/ch/epfl/bluebrain/nexus/kg/search/QueryBuilder.scala
```diff
@@ -7,12 +7,12 @@ import io.circe.Json
 object QueryBuilder {
 
   private def baseQuery(filterTerms: List[Json]): Json =
-    Json.obj(
-      "query" -> Json.obj(
-        "bool" -> Json.obj(
-          "filter" -> Json.arr(filterTerms: _*)
-        )
-      ))
+    Json.obj("sort" -> Json.arr(Json.fromString("@id")),
```
umbreak: Do we want to sort by id? is a it useless since the id's are mostly randomly generated

wwajerowicz: We need to have some sorting, so that the results are consistent. We could also sort  by time created/updated by default, but I'm not sure which one is better.

umbreak: maybe time created by default? Seems like a reasonable one?

wwajerowicz: 👍

PR #460, opened 2018-08-14
Added TopQuadrant Validator

umbreak:


build.sbt
```diff
@@ -123,7 +123,7 @@ lazy val kg = project
       serviceKafka,
       serviceKamon,
       serviceHttp,
-      wesoSchema,
+      topQuadrantShacl,
```
bogdanromanx: should also remove `resolvers  += Resolver.bintrayRepo("bogdanromanx", "maven")` as that's no longer required.

src/main/resources/logback.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
-
+    <statusListener class="ch.qos.logback.core.status.NopStatusListener" />
```
bogdanromanx: what does this do?

src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Rejection.scala
```diff
@@ -137,7 +137,7 @@ object Rejection {
     * @param schema a reference to the schema
     * @param report the validation report
     */
-  final case class InvalidResource(schema: Ref, report: ValidationReport)
+  final case class InvalidResource(schema: Ref, report: Json)
```
bogdanromanx: could we preserve the validation report and handle the json conversion at the routes level?

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/ResourceRoutes.scala
```diff
@@ -208,7 +203,7 @@ class ResourceRoutes(resources: Resources[Task])(implicit cache: DistributedCach
             }
 
           case `viewSchemaUri` =>
-            trace("listResolvers") {
+            trace("listViews") {
```
bogdanromanx: :metal:

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/ResourceRoutes.scala
```diff
@@ -266,7 +257,7 @@ class ResourceRoutes(resources: Resources[Task])(implicit cache: DistributedCach
       }
     } ~
       // add tag to resource
-      (evalBool(withTag) & pathPrefix("tags") & projectNotDeprecated) {
```
bogdanromanx: :metal:

src/main/resources/logback.xml
umbreak: https://stackoverflow.com/questions/3257154/how-to-prevent-logback-from-outputting-its-own-status-at-the-start-of-every-log

PR #462, opened 2018-08-14
Fix attachment handling.

wwajerowicz:


src/main/resources/app.conf
```diff
@@ -59,7 +59,7 @@ app {
 
   # Instance attachments specific settings
   attachments {
-    volume = "file:///tmp"
+    volume = "file:///tmp/" # ALWAYS INCLUDE THE TRAILING SLASH
```
bogdanromanx: we should normalize the path and handle with or without a trailing slash... this is easily forgotten and to be honest it should not be required.

src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/attachment/AttachmentStore.scala
```diff
@@ -177,7 +177,7 @@ object AttachmentStore {
           val relIri = Iri.relative(s"${id.parent.id}/${uuid.takeWhile(_ != '-').mkString("/")}/$uuid")
           EitherT.fromEither[F](relIri.left.map[Rejection](Unexpected).flatMap { relative =>
             Try {
-              val attachmentPath = new File(new File(base.show), relative.show).toPath
+              val attachmentPath = new File(new File(new URI(base.show)), relative.show).toPath
```
bogdanromanx: maybe we should do `new File(base.path.show, relative.show).toPath` instead; I don't think query params or fragments have a place in the binary location.

wwajerowicz: I agree, I think using `IRI`s here unnecessarily complicates things. We should just stick with `Path`s.

PR #469, opened 2018-08-16
Update nexus-admin and nexus-rdf versions

hygt:


src/test/resources/view/elasticview-wrong-2.json
```diff
@@ -1,9 +1,9 @@
 {
   "@id": "http://example.com/id",
   "@type": [
-    "View",
-    "ElasticView",
-    "Alpha"
+    "nxv:View",
```
umbreak: This is not needed, since where is used the [resolver context is applied](https://github.com/BlueBrain/nexus-kg/blob/master/src/test/scala/ch/epfl/bluebrain/nexus/kg/resources/ResourcesSpec.scala#L79) and that context has the neccessary alias.

src/test/resources/view/elasticview-wrong-2.json
```diff
@@ -1,9 +1,9 @@
...
+    "nxv:ElasticView",
```
umbreak: Same

src/test/resources/view/elasticview-wrong.json
```diff
@@ -1,9 +1,9 @@
 {
   "@id": "http://example.com/id",
   "@type": [
-    "View",
-    "ElasticView",
-    "Alpha"
+    "nxv:View",
```
umbreak: Same

src/test/resources/view/elasticview.json
```diff
@@ -1,9 +1,9 @@
 {
   "@id": "http://example.com/id",
   "@type": [
-    "View",
-    "ElasticView",
-    "Alpha"
+    "nxv:View",
```
umbreak: Same

src/test/resources/view/sparqlview-wrong.json
```diff
@@ -1,7 +1,7 @@
 {
   "@id": "http://example.com/id",
   "@type": [
-    "View",
-    "SparqlView"
+    "nxv:View",
```
umbreak: Same

PR #478, opened 2018-08-22
Checked the correctness of the elasticsearch mappings

### LINKED ISSUE #471: Creating ElasticSearch views with incorrect mappings should return 400 Bad Request
When creating an ElasticSearch view, the view gets created even when the mappings for the ES index are incorrect. The user has no way of knowing that the created view is invalid.
On top of that, KG will infinitely try to recreate the index and spam the logs.

umbreak:
Fixes #471

### DIFF
--- a/src/main/scala/ch/epfl/bluebrain/nexus/kg/resolve/Resolver.scala
+++ b/src/main/scala/ch/epfl/bluebrain/nexus/kg/resolve/Resolver.scala
@@ -2,19 +2,27 @@ package ch.epfl.bluebrain.nexus.kg.resolve
 
 import cats.instances.all._
 import cats.syntax.all._
+import ch.epfl.bluebrain.nexus.commons.http.syntax.circe._
 import ch.epfl.bluebrain.nexus.iam.client.types.Identity
 import ch.epfl.bluebrain.nexus.iam.client.types.Identity._
-import ch.epfl.bluebrain.nexus.kg.{DeprecatedId, RevisionedId}
+import ch.epfl.bluebrain.nexus.kg.config.Contexts.resolverCtxUri
 import ch.epfl.bluebrain.nexus.kg.config.Vocabulary._
-import ch.epfl.bluebrain.nexus.kg.resources.{AccountRef, ProjectRef, ResourceV}
+import ch.epfl.bluebrain.nexus.kg.resolve.ResolverEncoder.resolverGraphEncoder
+import ch.epfl.bluebrain.nexus.kg.resources._
+import ch.epfl.bluebrain.nexus.kg.resources.syntax._
+import ch.epfl.bluebrain.nexus.kg.{DeprecatedId, RevisionedId}
 import ch.epfl.bluebrain.nexus.rdf.Graph._
 import ch.epfl.bluebrain.nexus.rdf.Iri.AbsoluteIri
 import ch.epfl.bluebrain.nexus.rdf.Vocabulary._
 import ch.epfl.bluebrain.nexus.rdf.cursor.GraphCursor
+import ch.epfl.bluebrain.nexus.rdf.encoder.GraphEncoder.GraphResult
 import ch.epfl.bluebrain.nexus.rdf.encoder.NodeEncoder.EncoderResult
 import ch.epfl.bluebrain.nexus.rdf.encoder.NodeEncoderError.IllegalConversion
+import ch.epfl.bluebrain.nexus.rdf.syntax.circe._
+import ch.epfl.bluebrain.nexus.rdf.syntax.circe.context._
 import ch.epfl.bluebrain.nexus.rdf.syntax.node._
 import ch.epfl.bluebrain.nexus.rdf.syntax.node.encoder._
+import io.circe.Json
 
 /**
   * Enumeration of Resolver types.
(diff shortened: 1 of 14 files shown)


src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/AdditionalValidation.scala
```diff
@@ -30,20 +39,55 @@ trait AdditionalValidation[F[_]] {
...
+        val verifyMapping: F[Either[Rejection, Value]] = {
+          val index = UUID.randomUUID().toString
+          value.graph.cursor(id.value).downField(nxv.mapping).focus.as[String].flatMap(parse) match {
+            case Left(_) =>
+              F.pure(Left(InvalidPayload(id.ref, "ElasticSearch mapping field not found or not convertible to Json")))
+            case Right(mapping) =>
+              elastic
+                .createIndex(index, mapping)
+                .flatMap[Either[Rejection, Value]] {
+                  case true  => elastic.deleteIndex(index).map(_ => Right(value))
```
bogdanromanx: do we need to delete the index? could we instead create the index with the correct id?

src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/AdditionalValidation.scala
```diff
@@ -30,20 +39,55 @@ trait AdditionalValidation[F[_]] {
...
+    (id: ResId, schema: Ref, types: Set[AbsoluteIri], value: Value) => EitherT.rightT(value)
+
+  /**
+    * Additional validation used for checking the correctness of the ElasticSearch mappings
+    *
+    * @tparam F the monadic effect type
+    * @return a new validation that passes whenever the provided mappings are compliant with the ElasticSearch mappings or
+    *         when the view is not an ElasticView
+    */
+  final def view[F[_]](implicit F: MonadError[F, Throwable], elastic: ElasticClient[F]): AdditionalValidation[F] =
```
bogdanromanx: I'd recommend instantiating a View using it's companion apply method, pattern match on the type and for `ElasticView` just create the index based on its `name` method. On failure -> reject, on success -> pass the validation.

src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/AdditionalValidation.scala
```diff
@@ -60,34 +48,35 @@ object AdditionalValidation {
...
-      } else
-        EitherT.rightT[F, Rejection](value)
+  final def view[F[_]](implicit F: MonadError[F, Throwable],
+                       elastic: ElasticClient[F],
+                       config: ElasticConfig): AdditionalValidation[F] = {
 
-    }
+    (id: ResId, schema: Ref, types: Set[AbsoluteIri], value: Value) =>
+      {
+        def attemptCreateIndex(es: ElasticView) =
```
bogdanromanx: this could also be defined in the companion object `ElasticView`

src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/AdditionalValidation.scala
```diff
@@ -60,34 +48,35 @@ object AdditionalValidation {
...
+          elastic
+            .createIndex(s"${config.indexPrefix}_${es.name}", es.mapping)
+            .map[Either[Rejection, Value]] {
+              case true  => Right(value)
+              case false => Left(Unexpected("View mapping validation could not be performed"))
+            }
+            .recoverWith {
+              case err: ElasticFailure => F.pure(Left(InvalidPayload(id.ref, err.body)))
+              case err =>
```
bogdanromanx: `case NonFatal(err) =>` to exclude runtime errors

src/main/scala/ch/epfl/bluebrain/nexus/kg/resolve/Resolver.scala
```diff
@@ -45,6 +53,23 @@ sealed trait Resolver extends Product with Serializable {
...
     */
   def priority: Int
+
+  /**
+    * Converts a resolver to a [[ResourceF.Value]]
+    *
+    * @param id            the id of the resource
+    * @param flattenCtxObj the flatten context object (not wrapped in the @context key)
+    */
+  def toResourceV(id: ResId, flattenCtxObj: Json): ResourceF.Value = {
```
bogdanromanx: rename method to `resourceValue` as `ResourceV` is really a `ResourceF` parameterized on the `Value` type (it contains a `Value`)
... or return a `ResourceV` if there's enough information

src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/syntax.scala
```diff
@@ -45,4 +46,22 @@ object syntax {
...
+
+  final implicit class GraphSyntaxMeta(private val g: Graph) extends AnyVal {
+
+    /**
+      * Removes the metadata triples from the graph centered on the provided subject ''id''
+      *
+      * @param id the subject
+      * @return a new [[Graph]] without the metadata triples
+      */
+    def removeMetadata(id: IriOrBNode): Graph =
```
bogdanromanx: I'm wondering if this should be instead defined in the `ResourceF` companion object as the metadata is really something controller by a Resource

umbreak: It is better used as part of the syntax. I can rename it to `removeResourceMetadata` if that makes it more clear, or make something explicit on the docs.

bogdanromanx: I'm not arguing about the syntax, I think it's a good idea to provide syntax for the function, but the syntax can call the function in the companion object; this way we keep things that are strongly related together instead of spread around in various parts of the codebase.

PR #490, opened 2018-08-29
v0 events migration with Kafka

hygt:


src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/MigrationIndexer.scala
```diff
@@ -0,0 +1,101 @@
...
+      case Left(e)    => log.error(s"Couldn't migrate v0 event '${event.id}'", e)
+    }.runAsync
+  }
+  // $COVERAGE-ON$
+
+  private[indexing] def process(event: v0.Event): EitherT[Task, Rejection, Resource] = {
+    implicit val author: Identity = event.meta.toIdentity
+    event match {
+      case InstanceCreated(id, _, meta, value) =>
+        repo.create(toId(id), resourceSchema, resourceTypes, value, meta.instant)
```
umbreak: The types are always going to be `Resource` and not the type on the payload?

src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/v0/MigrationIndexer.scala
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+
+  private def toId(id: String): Id[ProjectRef] = Id(projectRef, base + id)
+
+  private def extractTypes(value: Json): Set[AbsoluteIri] = {
+    val types = value.asObject
+      .flatMap(_.apply("@type"))
+      .map(tpe => tpe.asString.toList ++ tpe.asArray.toList.flatMap(_.flatMap(_.asString))) match {
+      case Some(tpes) => tpes.flatMap(Iri.absolute(_).toSeq)
```
umbreak: Is not enough to convert to Iri.absolute. Things like `a:b` can be converted to an absolute iri surprisingly. You should do something like:
```scala
tpes.flatMap(Iri.url(_).toSeq.filter(_.scheme.value.matches("http(s?)")))
```

src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Repo.scala
```diff
@@ -91,31 +96,51 @@ class Repo[F[_]: Monad](agg: Agg[F], clock: Clock, toIdentifier: ResId => String
...
       if (res.deprecated) EitherT.leftT(IsDeprecated(id.ref))
       else if (res.rev != rev) EitherT.leftT(IncorrectRev(id.ref, rev))
       else
         store.save(id, attach, source).flatMap { attr =>
-          evaluate(id, AddAttachment(id, rev, attr, clock.instant(), identity))
+          evaluate(id, AddAttachment(id, rev, attr, instant, identity))
         }
     }
 
+  /**
```
umbreak: can you add a deprecated on the docs here? We should remember to delete this as soon as we don't need v0 migrations and this is not on a v0 package so we might forget

PR #491, opened 2018-08-29
Proxy ES responses.

wwajerowicz:


src/main/scala/ch/epfl/bluebrain/nexus/kg/marshallers/ExceptionHandling.scala
```diff
@@ -23,7 +25,8 @@ object ExceptionHandling {
     */
   final def apply(): ExceptionHandler =
     ExceptionHandler {
-      case rej: Rejection => complete(rej)
+      case rej: Rejection                   => complete(rej)
+      case ElasticClientError(status, body) => complete(status -> parse(body))
```
umbreak: You can also use the more generic case
```scala
case f: ElasticFailure => complete(status -> parse(f.body))
```

umbreak: Make it return a `Rejection`: 
```scala
complete(parse.left.map[Rejection](Unexpected("ElasticSearch response '$body'")))
```

src/test/scala/ch/epfl/bluebrain/nexus/kg/routes/ResourceRoutesSpec.scala
```diff
@@ -541,21 +542,36 @@ class ResourceRoutesSpec
...
-      Post(s"/v1/views/$account/$project/nxv:defaultElasticIndex/_search?size=23&other=value", query) ~> addCredentials(
+        elastic.searchRaw(mEq(query), mEq(Set(s"kg_${defaultEsView.name}")), mEq(Uri.Query(Map("other" -> "value"))))(
+          any[HttpClient[Task, Json]]()))
+        .thenReturn(Task.pure(esResponse))
+
+      Post(s"/v1/views/$account/$project/nxv:defaultElasticIndex/_search?other=value", query) ~> addCredentials(
         oauthToken) ~> routes ~> check {
         status shouldEqual StatusCodes.OK
-        responseAs[Json] shouldEqual Json.arr(result1, result2)
+        responseAs[Json] shouldEqual esResponse
```
umbreak: can you also add a test for when the body is not a Json, so the parser with return a `Left(Unexpected(...))`

src/main/scala/ch/epfl/bluebrain/nexus/kg/marshallers/ExceptionHandling.scala
wwajerowicz: @umbreak We probably don't want to forward ES server errors, so I don't think using the generic class/trait makes sense here.

PR #506, opened 2018-09-03
Fix content type on SPARQL endpoint.

wwajerowicz:


src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/ResourcesRoutes.scala
```diff
@@ -99,12 +102,13 @@ class ResourcesRoutes(resources: Resources[Task])(implicit cache: DistributedCac
   private def view(implicit token: Option[AuthToken]): Route = {
     val resourceRead = Permissions(Permission("views/read"), Permission("views/manage"))
 
+    implicit val um = PredefinedFromEntityUnmarshallers.stringUnmarshaller.forContentTypes(RdfMediaTypes.`application/sparql-query`, MediaTypes.`text/plain`)
```
umbreak: why don't we move than to the package `ch.epfl.bluebrain.nexus.kg.marshallers` ?

PR #514, opened 2018-09-10
Refactor indexing and DistributedCache.

wwajerowicz:


src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/Indexing.scala
```diff
@@ -46,95 +45,64 @@ private class Indexing(resources: Resources[Task], cache: DistributedCache[Task]
...
-            .addAccount(AccountRef(uuid), Account(org.name, rev, label, deprecated = false, uuid), updateRev = true)
-            .flatMap(processAccount(AccountRef(uuid)))
+            .addAccount(AccountRef(uuid), Account(org.name, rev, label, deprecated = false, uuid))
         case OrganizationDeprecated(_, uuid, rev, _) =>
-          cache.deprecateAccount(AccountRef(uuid), rev).flatMap(processAccount(AccountRef(uuid)))
+          cache.deprecateAccount(AccountRef(uuid), rev)
         case ProjectCreated(_, label, uuid, orgUUid, rev, _, proj) =>
-          cache
-            .addProject(
+          for {
```
umbreak: is this correct? Isn't supposed to not continue attempting to cache a view if the cache project return false?
Same for 
```scala
coordinator ! Msg(AccountRef(orgUUid), ProjectRef(uuid))
```
This should be called only when the project cache returns true, the view cache returns true and the resolver cache returns true

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/DistributedCache.scala
```diff
@@ -270,87 +258,101 @@ object DistributedCache {
...
+        _          <- updateProject(ref, proj)
+        _          <- addProjectToAccount(ref, accountRef)
+        _          <- addAccountRef(ref, accountRef, proj.rev)
+        accountSeg <- accountSegment(accountRef)
+        _          <- updateProjectLabelToUuid(ref, accountRef, proj, accountSeg)
+      } yield ()
     }
 
     /**
       * @return true if the project ref is already present so that we can chain the call during an update
```
umbreak: update docs

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/DistributedCache.scala
```diff
@@ -472,44 +480,38 @@ object DistributedCache {
...
     }
 
-    private def handleBooleanUpdate(timeoutMsg: String): PartialFunction[Any, Future[Boolean]] = {
+    private def handleUpdate(action: String): PartialFunction[Any, Future[Unit]] = {
       case UpdateSuccess(LWWRegisterKey(_), _) =>
-        Future.successful(true)
+        Future.successful(())
       case UpdateTimeout(LWWRegisterKey(_), _) =>
-        Future.failed(OperationTimedOut(timeoutMsg))
+        Future.failed(OperationTimedOut(s"Distributed cache update timed out while performing action '$action'"))
```
umbreak:
```scala
Future.failed(new RetriableErr(s"Distributed cache update timed out while performing action '$action'"))
```

PR #535, opened 2018-09-24
Fixed too many calls to admin/iam service

### LINKED ISSUE #534: Too many calls to ADMIN/IAM service
We are performing too many calls to fetch the ACLs. Every time we attempt to resolve (contexts or `ow:imports`) and if we have configured the project as `InAccount` or `CrossProject` resolver, we will make calls to ADMIN/IAM to fetch ACLs. For every single project defined in `CrossProject` and for every context or import.

As an example, [wholecellpatchedclamp](https://github.com/INCF/neuroshapes/blob/master/modules/experiment/src/main/resources/schemas/neurosciencegraph/experiment/wholecellpatchclamp/v0.1.0.json) has 18 `owl:imports`. Depending on how the organizations and projects are structured in the platform, that could be translated into a lot of calls to IAM.

umbreak:
Fixes #534

### DIFF
--- a/src/main/resources/app.conf
+++ b/src/main/resources/app.conf
@@ -77,7 +77,12 @@ app {
     base-uri = "http://localhost:8080/iam"
     base-uri = ${?IAM_BASEURI}
 
+    # The service account token to execute calls to IAM
     service-account-token = ${?IAM_SA_TOKEN}
+
+    # The maximum tolerated inactivity period after which the cached ACLs will be refreshed
+    cache-timeout = 30 seconds
+    cache-timeout = ${?IAM_CACHE_TIMEOUT}
   }
 
   # Sparql endpoint specific settings
--- a/src/main/resources/app.conf
+++ b/src/main/resources/app.conf
@@ -81,8 +81,8 @@ app {
     service-account-token = ${?IAM_SA_TOKEN}
 
     # The maximum tolerated inactivity period after which the cached ACLs will be refreshed
-    cache-timeout = 30 seconds
-    cache-timeout = ${?IAM_CACHE_TIMEOUT}
+    cache-refresh-interval = 30 seconds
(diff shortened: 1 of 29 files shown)


src/main/scala/ch/epfl/bluebrain/nexus/kg/acls/AclsActor.scala
```diff
@@ -0,0 +1,103 @@
...
+  // $COVERAGE-OFF$
+  final def start(name: String)(implicit client: IamClient[Task], iamConfig: IamConfig, as: ActorSystem): ActorRef = {
+    val props = ClusterSingletonManager.props(Props(new AclsActor(client)),
+                                              terminationMessage = Stop,
+                                              settings = ClusterSingletonManagerSettings(as))
+    val singletonManager = as.actorOf(props, name)
+    as.actorOf(
+      ClusterSingletonProxy.props(singletonManagerPath = singletonManager.path.toStringWithoutAddress,
+                                  settings = ClusterSingletonProxySettings(as)),
+      name = s"${name}Proxy"
```
hygt: So it's going to be called `iam-aclsProxy`, that's a bit weird. And given that we creating only one, I don't see why we need name as a parameter.

src/main/resources/app.conf
```diff
@@ -77,7 +77,12 @@ app {
     base-uri = "http://localhost:8080/iam"
     base-uri = ${?IAM_BASEURI}
 
+    # The service account token to execute calls to IAM
     service-account-token = ${?IAM_SA_TOKEN}
+
+    # The maximum tolerated inactivity period after which the cached ACLs will be refreshed
+    cache-timeout = 30 seconds
```
wwajerowicz: I think `cache-ttl`, `cache-expiry` or something similar  would be a better name for this, `cache-timeout` sounds more more like a timeout for retrieving something from cache.

PR #548, opened 2018-10-04
Bump several dependencies versions

hygt:


src/main/resources/akka.conf
```diff
@@ -29,6 +29,9 @@ akka {
 
       # Log implicitly registered classes. Useful, if you want to know all classes which are serialized.
       implicit-registration-logging = true
+
+      # enable compression
+      post-serialization-transformations="lz4"
```
bogdanromanx: formatting + add an env variable "KRYO_COMPRESSION" to be able to change the conf

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/ResourcesRoutes.scala
```diff
@@ -119,7 +119,7 @@ class ResourcesRoutes(resources: Resources[Task])(implicit cache: DistributedCac
 
     def search(implicit acls: FullAccessControlList, caller: Caller, wrapped: LabeledProject) =
       (pathPrefix(IdSegment / "sparql") & post & entity(as[String]) & pathEndOrSingleSlash) { (id, query) =>
-        (identity(caller) & hasPermission(resourceRead)) { implicit ident =>
+        (identity(caller) & hasPermission(resourceRead)) { _ =>
```
bogdanromanx: is the identity still propagated if you remove the implicit ?

hygt: The identity is not used here, hence the compiler warning.

PR #577, opened 2019-01-17
Added default resolvers and views to the primary store

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/80

### DIFF
--- a/src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/Indexing.scala
+++ b/src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/Indexing.scala
@@ -17,6 +17,7 @@ import ch.epfl.bluebrain.nexus.kg.config.AppConfig
 import ch.epfl.bluebrain.nexus.kg.config.AppConfig._
 import ch.epfl.bluebrain.nexus.kg.config.Contexts._
 import ch.epfl.bluebrain.nexus.kg.config.Schemas._
+import ch.epfl.bluebrain.nexus.kg.config.Vocabulary._
 import ch.epfl.bluebrain.nexus.kg.indexing.View.{ElasticView, SparqlView}
 import ch.epfl.bluebrain.nexus.kg.indexing.ViewEncoder._
 import ch.epfl.bluebrain.nexus.kg.resolve.Resolver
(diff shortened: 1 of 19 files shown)


src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/Indexing.scala
```diff
@@ -29,38 +43,29 @@ import scala.concurrent.Future
...
-//  private val sparqlUUID  = UUID.fromString("d88b71d2-b8a4-4744-bf22-2d99ef5bd26b")
+  private val logger                                          = Logger[this.type]
+  private val consumerSettings                                = ConsumerSettings(as, new StringDeserializer, new StringDeserializer)
+  private implicit val validation: AdditionalValidation[Task] = AdditionalValidation.pass
 
-//  private val defaultEsMapping =
-//    jsonContentOf("/elastic/mapping.json", Map(Pattern.quote("{{docType}}") -> config.elastic.docType))
+  private def asJson(view: View): Json =
+    view
+      .asJson(viewCtx.appendContextOf(resourceCtx))
```
bogdanromanx: why append the context and then remove it?

src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/Indexing.scala
```diff
@@ -69,7 +74,15 @@ private class Indexing(resources: Resources[Task], cache: Caches[Task], coordina
           // format: off
           val project = Project(config.http.projectsIri + label, label, orgLabel, desc, base, vocab, am, uuid, orgUuid, 1L, deprecated = false, instant, subject.id, instant, subject.id)
           // format: on
-          cache.project.replace(project) *> coordinator.start(project)
+          implicit val s         = subject
+          val elasticView: View  = ElasticView.default(project.ref)
+          val sparqlView: View   = SparqlView.default(project.ref)
+          val resolver: Resolver = InProjectResolver.default(project.ref)
+          cache.project.replace(project) *>
```
bogdanromanx: can we wrap this block in a `RetryStrategy[F]`? otherwise if one operation fails the coordinator won't start...

src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/Indexing.scala
```diff
@@ -69,7 +74,15 @@ private class Indexing(resources: Resources[Task], cache: Caches[Task], coordina
           // format: off
           val project = Project(config.http.projectsIri + label, label, orgLabel, desc, base, vocab, am, uuid, orgUuid, 1L, deprecated = false, instant, subject.id, instant, subject.id)
           // format: on
-          cache.project.replace(project) *> coordinator.start(project)
+          implicit val s         = subject
```
bogdanromanx: this is quite nice! :) the subject propagates across services!

src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/Indexing.scala
umbreak: In kg we have a conversion from those models to a Graph and then to Json. That's what the `asJson` does. Now, once we have the json, I don't want to store the whole context into the primary store, just with the link to the context is enough. That's why.

bogdanromanx: I see, it's just a code shortcut, ok.

PR #578, opened 2019-01-18
Changed Kafka client for SSE

umbreak:


src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/Indexing.scala
```diff
@@ -110,10 +117,19 @@ private class Indexing(resources: Resources[Task], cache: Caches[Task], coordina
         case ProjectDeprecated(uuid, rev, _, _) =>
           cache.project.deprecate(ProjectRef(uuid), rev) *> coordinator.stop(ProjectRef(uuid))
       }
-      update.runToFuture
     }
 
-    KafkaConsumer.start(consumerSettings, index, config.kafka.adminTopic, "admin-events", committable = false, None)
+    EventSource((config.admin.baseUri + "projects" + "events").toAkkaUri, send, None, 1 second)
```
bogdanromanx: the this needs to be `config.admin.baseUri + "events"`

build.sbt
```diff
@@ -115,6 +117,7 @@ lazy val kg = project
       akkaStream,
       akkaSlf4j,
       akkaCluster,
+      alpakkaSSE,
```
bogdanromanx: also remove the `serviceKafka` dependency

src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/Indexing.scala
```diff
@@ -73,11 +77,14 @@ private class Indexing(resources: Resources[Task], cache: Caches[Task], coordina
     case Right(value)                => Right(value)
   }
 
-  def startKafkaStream(): Unit = {
+  private def send(request: HttpRequest): Future[HttpResponse] =
```
bogdanromanx: need to update the request to put an `OAuth2BearerToken` to have access to the the stream

bogdanromanx: and also to log an error if the we can a 401 or 403

PR #581, opened 2019-01-18
Add routes to support underscore (any schema)

hygt:


src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/CommonRoutes.scala
```diff
@@ -38,34 +38,43 @@ private[routes] abstract class CommonRoutes(
...
 
-  implicit def additional: AdditionalValidation[Task] = AdditionalValidation.pass
+  private[routes] val simultaneousParamsRejection: AkkaRejection =
+    validationRejection("'rev' and 'tag' query parameters cannot be present simultaneously.")
+
+  protected val readPermission: Set[Permission] = Set(Permission.unsafe("resources/read"))
+
+  protected val writePermission: Set[Permission] =
+    if (prefix == "views") Set(Permission.unsafe("views/write"))
+    else Set(Permission.unsafe("resources/write"))
```
umbreak: why should we have only `views/write` and `resources/write` ?
schema, resolvers and the others might still be interesting

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/CommonRoutes.scala
```diff
@@ -106,12 +115,23 @@ private[routes] abstract class CommonRoutes(
...
 
-  def list(schema: Ref): Route =
-    (get & paginated & searchParams & hasPermission(resourceRead) & pathEndOrSingleSlash) { (pagination, params) =>
+  def list(schemaOpt: Option[Ref]): Route =
+    (get & paginated & searchParams & hasPermission(readPermission) & pathEndOrSingleSlash) { (pagination, params) =>
       trace(s"list$resourceName") {
-        val defaultView = viewCache.getBy[ElasticView](project.ref, nxv.defaultElasticIndex.value)
-        complete(
-          defaultView.flatMap(v => resources.list(v, params.copy(schema = Some(schema.iri)), pagination)).runToFuture)
+        schemaOpt match {
```
umbreak:
```scala
val schema = schemaOpt.getOrElse(params.schema)
complete(
viewCache.getBy[ElasticView](project.ref, nxv.defaultElasticIndex.value)
   .flatMap(v => resources.list(v, params.copy(schema = Some(schema)), pagination))
```

PR #586, opened 2019-01-21
Added context for default view/resolver ids

umbreak:


src/main/resources/contexts/resolver-context.json
```diff
@@ -39,5 +39,9 @@
     "subject": "nxv:subject",
     "group": "nxv:group"
   },
+  "account": {
```
bogdanromanx: I think this needs a better name... maybe "sameProject"? @MFSY what do you think it's a good alias for the default resolver for contexts and schemas (the one that looks up resources in the same project)?

MFSY: I think account should be nested within the @context object.

MFSY: I have to say that I'm a bit confused. But if I understand correctly:

* "Resolver"(mapping to "nxv:Resolver"): is a generic type for all resolvers
* "CrossProject" (mapping to "nxv:CrossProject"): is the type for resolvers that can look at many projects
* "InProject" (mapping to "nxv:InProject"): is the type for resolvers looking at the current project
* "account" (mapping to nxv:defaultResolver): is the type for resolvers looking at the current  organization (or account)

Or is it that InProject and Crossproject has nothing to do with resolvers ?

If the understanding above is correct then I suggest:

* to keep sync the alias name with the fragment in the mapped URI or CURIE
* to suffix every resolver type with "Resolver" (like CrossProjectResolver)
* To use "InAccountResolver" (mapping to nxv:InAccountResolver) as the type for resolvers looking at the current organization (or account)

PR #593, opened 2019-01-23
Fixed underscore routes

umbreak:
Changed order of directives in routes. The order must be as follows:
1. HTTP path
2. HTTP verb
3. HTTP query params (if present)
4. acls check
5. entity unmarshalling (entity(as[Json]))

### DIFF
--- a/src/main/scala/ch/epfl/bluebrain/nexus/kg/directives/QueryDirectives.scala
+++ b/src/main/scala/ch/epfl/bluebrain/nexus/kg/directives/QueryDirectives.scala
@@ -3,7 +3,7 @@ package ch.epfl.bluebrain.nexus.kg.directives
 import akka.http.scaladsl.common.{NameOptionReceptacle, NameReceptacle}
 import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.directives.ParameterDirectives.ParamDefAux
-import akka.http.scaladsl.server.{Directive0, Directive1, ValidationRejection}
+import akka.http.scaladsl.server.{Directive0, Directive1, MalformedQueryParamRejection}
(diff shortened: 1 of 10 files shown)


src/main/scala/ch/epfl/bluebrain/nexus/kg/directives/QueryDirectives.scala
```diff
@@ -18,6 +20,20 @@ object QueryDirectives {
...
+  /**
+    * This directive passes when the query parameter specified is not present
+    *
+    * @param param the parameter
+    * @tparam A the type of the parameter
+    */
+  def notParameter[A](param: NameReceptacle[A])(
+      implicit paramAux: ParamDefAux[NameOptionReceptacle[A], Directive1[Option[A]]]): Directive0 = {
+    parameter(param.?).flatMap {
+      case Some(_) => reject(ValidationRejection(s"the provided query parameter '${param.name}' should not be present"))
```
bogdanromanx: please use `MalformedQueryParamRejection` instead

src/main/scala/ch/epfl/bluebrain/nexus/kg/directives/QueryDirectives.scala
```diff
@@ -18,6 +20,22 @@ object QueryDirectives {
       case (from, size) => Pagination(from.max(0), size.max(1).min(config.sizeLimit))
     }
 
+  /**
+    * This directive passes when the query parameter specified is not present
+    *
+    * @param param the parameter
+    * @tparam A the type of the parameter
+    */
+  def notParameter[A](param: NameReceptacle[A])(
```
hygt: `noParameter` maybe?

PR #595, opened 2019-01-23
Refactored Exception and Rejection handling

bogdanromanx:


src/main/scala/ch/epfl/bluebrain/nexus/kg/KgError.scala
```diff
@@ -0,0 +1,97 @@
...
+    * Generic wrapper for kg errors that should not be exposed to clients.
+    *
+    * @param reason the underlying error reason
+    */
+  final case class InternalError(reason: String) extends KgError(reason)
+
+  /**
+    * Signals that the requested resource was not found
+    */
+  final case class NotFound(ref: Option[String] = None) extends KgError("The requested resource could not be found.")
```
wwajerowicz: It would be useful to have the `ref` somewhere in the error message.

wwajerowicz: what is the use case of having an `Option` here? when would it be `None`?

src/main/scala/ch/epfl/bluebrain/nexus/kg/directives/ProjectDirectives.scala
```diff
@@ -70,8 +64,6 @@ object ProjectDirectives {
     * @return pass when the project is not deprecated, rejects when project is deprecated
     */
   def projectNotDeprecated(implicit proj: Project): Directive0 =
-    if (proj.deprecated)
-      reject(CustomAuthRejection(ProjectIsDeprecated(ProjectLabel(proj.organizationLabel, proj.label))))
-    else
-      pass
+    if (proj.deprecated) failWith(ProjectIsDeprecated(ProjectLabel(proj.organizationLabel, proj.label)))
```
umbreak: import on the top `ch.epfl.bluebrain.nexus.kg.resources.syntax._`
and then:
```scala
ProjectIsDeprecated(proj.projectLabel)
```

src/main/scala/ch/epfl/bluebrain/nexus/kg/KgError.scala
```diff
@@ -0,0 +1,97 @@
...
+    */
+  final case class OperationTimedOut(override val msg: String) extends KgError(msg)
+
+  /**
+    * Signals the impossibility to resolve the project reference for project labels.
+    *
+    * @param label the project label
+    */
+  final case class ProjectNotFound(label: ProjectLabel)
+      extends KgError(s"Project reference for label '${label.show}' not found.")
```
wwajerowicz: This message is a bit confusing, perhaps `"Project  '${label.show}' not found."` would be simpler?

src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Rejection.scala
```diff
@@ -29,23 +38,23 @@ object Rejection {
...
   /**
     * Signals an attempt to perform a request with an invalid payload.
     *
     * @param ref a reference to the resource
-    * @param reason the human readable reason for the rejection
+    * @param details the human readable reason for the rejection
     */
-  final case class InvalidPayload(ref: Ref, reason: String)
-      extends Rejection(s"Resource '${ref.show}' with invalid payload due to '$reason'.")
+  final case class InvalidPayload(ref: Ref, details: String)
```
umbreak: In admin we have something like `InvalidProjectFormat`. Maybe we could name it `InvalidResourceFormat` ?

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/CommonRoutes.scala
```diff
@@ -76,7 +75,7 @@ private[routes] abstract class CommonRoutes(
   def routes: Route
 
   def create(schema: Ref): Route =
-    (post & noParameter('rev.as[Long]) & projectNotDeprecated & pathEndOrSingleSlash & hasPermission(writePermission)) {
+    (post & noParameter('rev.as[Long]) & projectNotDeprecated & pathEndOrSingleSlash & hasPermissions(writePermission)) {
```
umbreak: `hasPermissions(writePermission)` seems like redundant. We can just rename
writePermission -> write
readPermission -> read

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/Routes.scala
```diff
@@ -1,54 +1,183 @@
...
+    )
+    val appInfoRoutes = AppInfoRoutes(config.description, healthStatusGroup).routes
+
+    wrap(
+      pathPrefix(config.http.prefix / Segment) { resourceSegment =>
+        extractToken {
+          implicit optToken =>
+            project.apply {
+              implicit project =>
+                (extractCallerAcls & extractCaller) { (acl, c) =>
```
umbreak: I still don't see the point why to make this call to IAM `extractCallerAcls` if we have this information into the cache. 
I think the real fix, which we should do later is to update the cache immediately subscribing to IAM SSE.

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/ViewRoutes.scala
```diff
@@ -73,7 +75,7 @@ class ViewRoutes private[routes] (resources: Resources[Task], acls: AccessContro
 
   private def sparql: Route =
     pathPrefix(IdSegment / "sparql") { id =>
-      (post & pathEndOrSingleSlash & hasPermission(queryPermission)) {
+      (post & pathEndOrSingleSlash & hasPermissions(queryPermission)) {
```
umbreak: rename queryPermission -> query ?

src/main/scala/ch/epfl/bluebrain/nexus/kg/marshallers/instances.scala
```diff
@@ -74,30 +78,16 @@ package object instances extends FailFastCirceSupport {
...
+  implicit class EitherTask[R <: Rejection, A](task: Task[Either[R, A]])(implicit s: Scheduler) {
+    def runWithStatus(code: StatusCode): Future[Either[R, (StatusCode, A)]] =
+      task.map(_.map(code -> _)).runToFuture
+  }
+
+  implicit class OptionTask[A](task: Task[Option[A]])(implicit s: Scheduler) {
+    def runNotFound: Future[A] =
+      task.flatMap {
+        case Some(a) => Task.pure(a)
+        case None    => Task.raiseError(KgError.NotFound())
```
umbreak: NotFound contains an optional ref just because of this call.
Maybe we could have a typeclass
```trait Referenced[A]{
   def apply(value: A): Ref 
}
```
And then have the signature here as `OptionTask[A: Referenced]`

src/main/scala/ch/epfl/bluebrain/nexus/kg/KgError.scala
bogdanromanx: the ref is added a a separate field, and the case for None happens when the the routes are terminated without knowing the exact path

PR #610, opened 2019-02-15
Expose project event stream via SSE.

wwajerowicz:


src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/EventRoutes.scala
```diff
@@ -0,0 +1,91 @@
...
+class EventRoutes(acls: AccessControlLists, caller: Caller)(implicit as: ActorSystem,
+                                                            project: Project,
+                                                            config: AppConfig) {
+
+  private val printer: Printer = Printer.noSpaces.copy(dropNullValues = true)
+
+  private val pq: EventsByTagQuery =
+    PersistenceQuery(as).readJournalFor[EventsByTagQuery](config.persistence.queryJournalPlugin)
+
+  private val resourcesRead: Set[Permission]    = Set(Permission.unsafe("resources/read"))
```
umbreak: can we call it just `private val read: Set[Permission]`, just to have the same as in CommonRoutes

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/EventRoutes.scala
```diff
@@ -0,0 +1,91 @@
...
+  private implicit val acl: AccessControlLists  = acls
+  private implicit val c: Caller                = caller
+  private implicit val iamConf: IamClientConfig = config.iam.iamClient
+
+  def routes: Route =
+    lastEventId { offset =>
```
umbreak: you could merge those two 
```scala
(lastEventId && hasPermissions(resourcesRead)) {offset => ...}
```
so that we avoid the .apply

src/test/resources/events/events.json
```diff
@@ -0,0 +1,90 @@
+[
+  {
+    "@context": "https://bluebrain.github.io/nexus/contexts/resource.json",
+    "@type": "Created",
+    "resourceId": "http://example.com/base/Created",
+    "schema": "http://example.com/base/schema",
```
bogdanromanx: constrainedBy?

src/test/resources/events/events.json
```diff
@@ -0,0 +1,90 @@
...
+    "attributes": {
+      "bytes": 47,
+      "digest": {
+        "algorithm": "SHA-256",
+        "value": "00ff4b34e3f3695c3abcdec61cba72c2238ed172ef34ae1196bfad6a4ec23dda"
+      },
+      "filename": "attachment.json",
+      "mediaType": "application/json"
+    },
+    "resourceId": "http://example.com/base/file",
```
bogdanromanx: resourceId before attributes... if it's not too much trouble

src/test/resources/events/events.json
```diff
@@ -0,0 +1,90 @@
...
+    "targetRev": 2,
+    "_project": "http://example.com/base/org/project",
+    "_rev": 4,
+    "_instant": "1970-01-01T00:00:00Z",
+    "_subject": "http://localhost:8080/v1/realms/myrealm/users/uuid"
+  },
+  {
+    "@context": "https://bluebrain.github.io/nexus/contexts/resource.json",
+    "@type": "FileCreated",
+    "attributes": {
```
bogdanromanx: this should be prefixed with _

bogdanromanx: and the inner keys

src/test/resources/events/events.json
```diff
@@ -0,0 +1,90 @@
...
+    ],
+    "_project": "http://example.com/base/org/project",
+    "_instant": "1970-01-01T00:00:00Z",
+    "_subject": "http://localhost:8080/v1/realms/myrealm/users/uuid"
+  },
+  {
+    "@context": "https://bluebrain.github.io/nexus/contexts/resource.json",
+    "@type": "Updated",
+    "resourceId": "http://example.com/base/created",
+    "source": {
```
bogdanromanx: prefix source with _

src/test/resources/events/events.json
```diff
@@ -0,0 +1,90 @@
...
+      "@type": "Updated"
+    },
+    "types": [
```
bogdanromanx: prefix types with _

src/test/resources/events/events.json
```diff
@@ -0,0 +1,90 @@
+[
+  {
+    "@context": "https://bluebrain.github.io/nexus/contexts/resource.json",
+    "@type": "Created",
+    "resourceId": "http://example.com/base/Created",
```
bogdanromanx: I'm wondering if we should use `@id` here or not; for realms we do use `@id`; we don't for orgs and projects as they define labels and uuids instead. Thoughts?

umbreak: I think @id in the events does not make sense if we follow the idea that an event is NOT the resource itself. But well It wouldn't be tragic if we would use it either...

PR #611, opened 2019-02-15
Added support for bulk indexing on Blazegraph

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/451

### DIFF
--- a/src/main/resources/contexts/resource-context.json
+++ b/src/main/resources/contexts/resource-context.json
@@ -1,46 +1,16 @@
 {
   "@context": {
     "nxv": "https://bluebrain.github.io/nexus/vocabulary/",
-    "Schema": {
-      "@id": "nxv:Schema",
-      "@type": "@id"
-    },
-    "File": {
-      "@id": "nxv:File",
-      "@type": "@id"
-    },
-    "Resolver": {
-      "@id": "nxv:Resolver",
-      "@type": "@id"
-    },
-    "InProject": {
-      "@id": "nxv:InProject",
-      "@type": "@id"
-    },
-    "CrossProject": {
-      "@id": "nxv:CrossProject",
-      "@type": "@id"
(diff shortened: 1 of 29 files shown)


src/main/resources/contexts/resource-context.json
```diff
@@ -1,46 +1,16 @@
...
-    },
-    "SparqlView": {
-      "@id": "nxv:SparqlView",
-      "@type": "@id"
-    },
-    "Alpha": {
-      "@id": "nxv:Alpha",
-      "@type": "@id"
-    },
+    "Schema": "nxv:Schema",
```
bogdanromanx: why this change?

src/main/scala/ch/epfl/bluebrain/nexus/kg/config/AppConfig.scala
```diff
@@ -189,6 +189,17 @@ object AppConfig {
...
   final case class IndexingConfig(batch: Int, batchTimeout: FiniteDuration, retry: RetryStrategyConfig)
 
+  /**
+    * The indexing configuration for the different indexers
+    *
+    * @param sparql        the indexing configuration applied to Sparql indexer
+    * @param elasticSearch the indexing configuration applied to Elastic Search indexer
+    * @param keyValueStore the indexing configuration applied to the Distributed Data store
+    */
+  final case class IndexingConfigCollection(sparql: IndexingConfig,
```
bogdanromanx: rename to `IndexingConfigs` ?

src/main/resources/contexts/resource-context.json
umbreak: Because it is not needed. Those aliases are for things like
```json
{
   "@type": "Schema"
}
```
So an alias is enough, as you can check with the playground.

PR #612, opened 2019-02-18
Expose global event stream over SSE.

wwajerowicz:


src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/GlobalEventRoutes.scala
```diff
@@ -0,0 +1,31 @@
...
+import ch.epfl.bluebrain.nexus.kg.directives.AuthDirectives._
+import ch.epfl.bluebrain.nexus.kg.persistence.TaggingAdapter
+
+class GlobalEventRoutes(acls: AccessControlLists, caller: Caller)(implicit as: ActorSystem, config: AppConfig)
+    extends EventCommonRoutes {
+
+  override val pq: EventsByTagQuery =
+    PersistenceQuery(as).readJournalFor[EventsByTagQuery](config.persistence.queryJournalPlugin)
+
+  private val read: Set[Permission]             = Set(Permission.unsafe("resources/read"), Permission.unsafe("events/read"))
```
bogdanromanx: the permission needs to be `events/read` as with the other services.

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/EventCommonRoutes.scala
```diff
@@ -0,0 +1,79 @@
...
+import io.circe.syntax._
+import io.circe.{Encoder, Printer}
+
+import scala.util.{Failure, Success, Try}
+import scala.concurrent.duration._
+
+/**
+  * Defines commons methods for event routes.
+  */
+trait EventCommonRoutes {
```
bogdanromanx: I would recommend making this an abstract class that takes the required fields in the constructor; this avoids possible initialization issues.

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/Routes.scala
```diff
@@ -163,27 +163,24 @@ object Routes {
...
-                    case "schemas"   => new SchemaRoutes(resources, acl, c).routes
-                    case "files"     => new FileRoutes(resources, acl, c).routes
-                    case "resources" => new ResourceRoutes(resources, acl, c).routes
-                    case "events"    => new EventRoutes(acl, c).routes
-                    case _           => reject()
-                  }
-                }
+    wrap(extractToken { implicit optToken =>
+      (extractCallerAcls & extractCaller) { (acl, c) =>
+        pathPrefix(config.http.prefix / Segment) {
```
bogdanromanx: I would recommend using `concat` instead of `~` to join routes together; it makes stuff easier to read.

src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/GlobalEventRoutes.scala
wwajerowicz: but wouldn't it make sense to allow both? if `resource/read` on `/org/project` gives you access to event stream for a project, shouldn't `resource/read` also give you access to full event stream?

bogdanromanx: I think we need to guard event streaming with a separate permission (which is why in iam and admin to get the entire event log you need `events/read`).

For v1.1 we can consider applying different permissions and change the routes to use different paths... but for now, even if it's just for consistency, we should use `events/read` for the global event log.

PR #617, opened 2019-02-26
Add basic trait for Storage

hygt:


src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/file/Store.scala
```diff
@@ -0,0 +1,67 @@
...
+import java.util.UUID
+
+import ch.epfl.bluebrain.nexus.kg.resources.{ProjectRef, ResId}
+import ch.epfl.bluebrain.nexus.kg.resources.file.File.{FileAttributes, FileDescription}
+import ch.epfl.bluebrain.nexus.rdf.Iri.AbsoluteIri
+
+/**
+  * Contract for different types of storage back-end.
+  */
+trait Store[F[_], In, Out] {
```
umbreak: Having `In` and `Out` directly here can be problematic because then the cache will have to be parametrized with those generic types. However the cache is going to have to deal with different `In` and `Out` values.
Why not having those directly in the `save` and `fetch` methods and having a typeclass which actually implements the logic?

 ```scala
def save[F[_], In, StoreType](id: ResId, fileDesc: FileDescription, source: In)(implicit ops: StoreOperations[F, StoreType, In]): F[FileAttributes] = ops.save(id, fileDesc, source)

def fetch[Out](fileMeta: FileAttributes)(implicit ops: StoreOperations[F, StoreType, In]): Out  = ops.fetch(fileMeta)
```

File: src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/file/Store.scala [hunk 2]
```diff
@@ -0,0 +1,67 @@
...
+
```
umbreak: Can we also add:
```scala
  /**
    * @return the instant when this store was recorded
    */
  def instant: Instant
```

PR #619, opened 2019-02-27
Initial file storage refactoring

hygt:


src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/file/FileStorage.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  private val logger: Logger = Logger[this.type]
+
+  /**
+    * Wraps both the absolute and relative information about the file location
+    *
+    * @param path     absolute Path where to find the file
+    * @param relative relative route to the file location
+    */
+  final case class Location(path: Path, relative: Path)
```
umbreak: we can make this private, since it is scoped on this method. And we can remove the docs then

PR #628, opened 2019-03-11
Added storage validation

umbreak:


src/test/scala/ch/epfl/bluebrain/nexus/kg/resources/AdditionalValidationSpec.scala
```diff
@@ -260,6 +266,25 @@ class AdditionalValidationSpec
         validation(id, schema, types, resource.value, 1L).value.accepted shouldEqual resource.value
       }
     }
+
+    "applied to storages" should {
```
bogdanromanx: add additional test cases for:
- when the volume is not writable
- when the volume is not a directory

src/main/scala/ch/epfl/bluebrain/nexus/kg/storage/DiskStorageOperations.scala
```diff
@@ -5,19 +5,33 @@ import java.nio.file.{Files, Path, Paths}
...
 import scala.util.control.NonFatal
 
 object DiskStorageOperations {
 
+  /**
+    * [[VerifyStorage]] implementation for [[DiskStorage]]
+    *
+    * @param storage the [[DiskStorage]]
+    */
+  final class Verify[F[_]](storage: DiskStorage)(implicit F: Monad[F]) extends VerifyStorage[F] {
```
bogdanromanx: verify extends verifystorage is a bit weird... abstract extends specific...

PR #632, opened 2019-03-19
Added AggregatedSparqlView

umbreak:
- Refactored configuration.


src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/View.scala
```diff
@@ -128,7 +136,36 @@ object View {
...
     */
-  sealed trait AggregateView extends View
+  sealed trait AggregateView[P] extends View {
+
+    /**
+      * @return the set of views that this view connects to when performing searches
+      */
+    def value: Set[ViewRef[P]]
+
+    def toValue[PP](newValue: Set[ViewRef[PP]]): AggregateView[PP] = this match {
```
bogdanromanx: scapegoat complains here... a `@silent` would do.

PR #633, opened 2019-03-19
Fix default S3 region selection

hygt:


src/main/scala/ch/epfl/bluebrain/nexus/kg/storage/Storage.scala
```diff
@@ -195,12 +195,14 @@ object Storage {
         case None => new AWSStaticCredentialsProvider(new AnonymousAWSCredentials)
       }
 
-      val regionProvider = region match {
-        case Some(reg) =>
-          new AwsRegionProvider {
-            val getRegion: String = reg
+      val regionProvider: AwsRegionProvider = new AwsRegionProvider {
```
bogdanromanx: does this mean the only supported region is us-east-1?

hygt: It means if you don't specify the region and use Amazon S3, we default to `us-east-1` (which is what  Amazon does when you access https://s3.amazonaws.com without a region). The region is used in the authentication signature and the Alpakka connector really expects a non-null string.

If you specify an endpoint that doesn't look like AWS, then it's most likely irrelevant, any string should do unless your S3 compatible expects a specific region string.

umbreak: We could also define the default region on `app.conf`, as we do for other default values?

bogdanromanx: Agreed on the comment to put the default in the config.

Also, as I can see from the code, the only possible region values are "us-east-1" or "no-region"; is that expected?

umbreak: @hygt explained me why is that by default. The region can and should be specified when creating an `S3Storage`. If it is not set, we default to what S3 defaults on its clients. So anybody using S3 clients have the same expectations when using nexus as well. It makes sense.

hygt: `region.getOrElse` ...

A default config doesn't make sense since the user would have no way to know and would get an inconsistent behavior compared to all S3 clients/SDK out there. If you don't specify a region then your requests go to the default endpoint.

bogdanromanx: makes sense, I've missed the `region.getOrElse`...

PR #641, opened 2019-03-27
Added surface API for each resource

umbreak:
- Views
- Resolvers
- Schemas
- Storages
- Files

Fixes https://github.com/BlueBrain/nexus/issues/584
Fixes https://github.com/BlueBrain/nexus/issues/518


.java-version
```diff
@@ -0,0 +1 @@
+oracle64-10.0.2
```
wwajerowicz: shouldn't we be using OpenJDK like in our Dockerimages?

umbreak: This file shouldn't even be there. I'll remove it

PR #644, opened 2019-04-02
Updated SSE API to expose event streams as a sub-resource

bogdanromanx:
Updated sourcing dependency.

### DIFF
--- a/src/main/scala/ch/epfl/bluebrain/nexus/kg/async/ProjectViewCoordinatorActor.scala
+++ b/src/main/scala/ch/epfl/bluebrain/nexus/kg/async/ProjectViewCoordinatorActor.scala
@@ -118,7 +118,7 @@ private abstract class ProjectViewCoordinatorActor(viewCache: ViewCache[Task])(i
         .plugin(config.persistence.queryJournalPlugin)
         .retry[ElasticSearchServerOrUnexpectedFailure](indexing.retry.retryStrategy)(elasticErrorMonadError)
         .batch(indexing.batch, indexing.batchTimeout)
-        .restart(value = false)
+        .restart(false)
         .init(Task.unit)
         .mapping[Event, Event](a => Task.pure(Some(a)))
         .index(_ => Task.unit)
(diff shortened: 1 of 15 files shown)


src/main/scala/ch/epfl/bluebrain/nexus/kg/async/ProjectViewCoordinatorActor.scala
```diff
@@ -110,15 +110,15 @@ private abstract class ProjectViewCoordinatorActor(viewCache: ViewCache[Task])(i
...
+      ProjectionConfig
         .builder[Task]
         .name(s"project-event-count-${project.uuid}")
         .tag(s"project=${project.uuid}")
         .plugin(config.persistence.queryJournalPlugin)
-        .retry[ElasticSearchServerOrUnexpectedFailure](indexing.retry.retryStrategy)
+        .retry[ElasticSearchServerOrUnexpectedFailure](indexing.retry.retryStrategy)(elasticErrorMonadError)
         .batch(indexing.batch, indexing.batchTimeout)
-        .restart(false)
+        .restart(value = false)
```
umbreak: even though IntellIj does not like to have booleans without specifying the parameter, in this case it doesn't make sense, since it is a builder. The information is already on the method (restart(true) or restart(false)).

src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/ResolverIndexer.scala
```diff
@@ -50,37 +46,52 @@ private class ResolverIndexerMapping[F[_]: Timer](resources: Resources[F])(impli
...
         .name("resolver-indexer")
         .tag(s"type=${nxv.Resolver.value.show}")
         .plugin(config.persistence.queryJournalPlugin)
-        .retry[KgError](indexing.retry.retryStrategy)
+        .retry[KgError](indexing.retry.retryStrategy)(kgErrorMonadError)
         .batch(indexing.batch, indexing.batchTimeout)
         .offset(Volatile)
         .mapping(mapper.apply)
-        .index(_.traverse(resolverCache.put) *> Task.unit)
+        .index(resolverList => resolverList.traverse(resolverCache.put) >> F.unit)
```
umbreak: Why did you make explicit the variable `resolverList` in this case and not in the other cases? I would say that either everywhere or nowhere (StorageIndexer.scala, ViewIndexer.scala, ...)

bogdanromanx: This was an artefact of my attempt to sort out the compilation errors which were caused by the imported monaderror.

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/ProjectViewCoordinatorActor.scala
bogdanromanx: agreed

PR #647, opened 2019-04-05
Create ES index when creating default ES view.

wwajerowicz:


src/main/scala/ch/epfl/bluebrain/nexus/kg/indexing/Indexing.scala
```diff
@@ -93,12 +95,20 @@ private class Indexing(
...
-      created.mapRetry(
-        createdOrExists,
-        InternalError(s"Couldn't create default ElasticSearch view for project '${project.ref}'"): KgError)
+      view
+        .createIndex[Task]
+        .recover {
+          case err =>
+            logger.error("There was a problem while creating index for default ElasticSearch view", err)
+            ()
+        } *> resources
```
umbreak: Use `>>` or `.flatMap` instead of `*>` since this latest does not have a requirement on which executes first.

PR #648, opened 2019-04-11
Add S3 linking operation

hygt:


src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Files.scala
```diff
@@ -67,6 +68,69 @@ class Files[F[_]: Effect: Timer](repo: Repo[F])(implicit config: AppConfig) {
...
+    *
+    * @param projectRef reference for the project in which the resource is going to be created.\
+    * @param base       base used to generate new ids
+    * @param storage    the storage where the file is going to be saved
+    * @param fileDesc   the file description metadata
+    * @param location   the file location URI
+    * @param rev        the last known resource revision
+    * @return either a rejection or the new resource representation in the F context
+    */
+  def updateLink(projectRef: ProjectRef,
```
umbreak: What is this used for? When updating a resource in nexus you always have the @id and you never autogenerate a new one.

src/main/scala/ch/epfl/bluebrain/nexus/kg/storage/package.scala
```diff
@@ -6,24 +6,28 @@ import java.security.MessageDigest
...
-  def digestSink(algorithm: String): Sink[ByteString, Future[MessageDigest]] =
-    Sink.fold(MessageDigest.getInstance(algorithm))((digest, currentBytes) => {
-      digest.update(currentBytes.asByteBuffer)
-      digest
-    })
+  def digestSink(algorithm: String): Sink[ByteString, Future[Digest]] = {
+    val digest = MessageDigest.getInstance(algorithm)
+    Flow[ByteString]
+      .map(bs => digest.update(bs.asByteBuffer))
+      .map(_ => Digest(digest.getAlgorithm, digest.digest.map("%02x".format(_)).mkString))
```
umbreak: Doesn't this perform the `digest.digest.map("%02x".format(_)` for every chunk on the Source ?
Also, `Sink.head`
> Materializes into a Future which completes with the first value arriving, after this the stream is canceled

which doesn't seem to be what we want to do here.

Why not:
```scala
def digestSink(algorithm: String): Sink[ByteString, Future[Digest]] =
      Sink
        .fold(MessageDigest.getInstance(algorithm))((digest, currentBytes: ByteString) => {
          digest.update(currentBytes.asByteBuffer)
          digest
        })
        .mapMaterializedValue(_.map(dig => Digest(dig.getAlgorithm, dig.digest().map("%02x".format(_)).mkString)))
```

PR #650, opened 2019-04-26
Add S3 linking and routes

hygt:


src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/file/File.scala
```diff
@@ -4,11 +4,78 @@ import java.util.UUID
...
+          .downField(nxv.filename)
+          .focus
+          .as[String]
+          .left
+          .map(_ => InvalidResourceFormat(id.ref, "Invalid 'filename' field."))
+        mediaType <- c
+          .downField(nxv.mediaType)
+          .focus
+          .as[String]
+          .left
```
umbreak: instead of .left.map(...) you can do `import ch.epfl.bluebrain.nexus.kg.resources.syntax._` and do `.toRejectionOnLeft(id.ref)`

src/main/scala/ch/epfl/bluebrain/nexus/kg/storage/S3StorageOperations.scala
```diff
@@ -134,4 +123,47 @@ object S3StorageOperations {
...
   }
 
+  final class Link[F[_]](storage: S3Storage)(implicit F: Effect[F], as: ActorSystem, config: StorageConfig)
+      extends LinkFile[F] {
+
+    private implicit val ec: ExecutionContext = as.dispatcher
+    private implicit val mt: Materializer     = ActorMaterializer()
+
+    override def apply(id: ResId, fileDesc: FileDescription, location: Uri): F[FileAttributes] =
+      getKey(storage.bucket, location.path) match {
```
umbreak: I guess here we should have a check that the `location` contains the storage endpoint

src/main/scala/ch/epfl/bluebrain/nexus/kg/storage/S3StorageOperations.scala
```diff
@@ -159,11 +158,25 @@ object S3StorageOperations {
...
   }
 
-  private def getKey(bucket: String, path: Path): Option[String] = path match {
-    case Slash(Segment(head, Slash(tail))) if head == bucket =>
-      Some(URLDecoder.decode(tail.toString, "UTF-8"))
-    case _ =>
-      logger.error(s"Error decoding key from S3 object URI '$path' in bucket '$bucket'")
-      None
+  private def getKey(storage: S3Storage, location: Uri): Option[String] = {
+    val path = storage.settings.endpoint match {
```
umbreak: you could use collect to make it shorter. Something like that should work:
```scala
val path = storage.settings.endpoint.collect {
   case Some(endpoint) if location.toString.startsWith(endpoint) || endpoint.startsWith("http://localhost") => location.path
   case None if location.authority.toString.endsWith("amazonaws.com") => location.path
}
```

hygt: `collect` won't work if the optional endpoint is empty

PR #599, opened 2019-01-25
Renamed Elastic* to ElasticSearch*

bogdanromanx:

PR #652, opened 2019-05-03
Added ExternalDiskStorge

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/599


src/test/scala/ch/epfl/bluebrain/nexus/kg/storage/ExternalDiskStorageOperationsSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+    implicit val token: Option[AuthToken] = Some(AuthToken(cred))
+    val path                              = Uri.Path(s"${genString()}/${genString()}")
+    // format: off
+    val storage = ExternalDiskStorage(ProjectRef(genUUID), genIri, 1L, false, false, "SHA-256", endpoint, Some(cred.encrypt), genString(), Permission.unsafe(genString()), Permission.unsafe(genString()))
+    val attributes = FileAttributes(s"$endpoint/${storage.folder}/$path", path, s"${genString()}.json", `application/json`, 12L, Digest("SHA-256", genString()))
+    // format: on
+  }
+
+  private def sourceInChunks(input: String): AkkaSource =
+    Source.fromIterator(() ⇒ input.grouped(10000).map(ByteString(_)))
```
hygt: Unicode arrow

src/test/scala/ch/epfl/bluebrain/nexus/kg/storage/ExternalDiskStorageOperationsSpec.scala
```diff
@@ -0,0 +1,104 @@
...
+      val verify = new ExternalDiskStorageOperations.Verify[IO](storage, client)
+      verify.apply
+        .rejected[String] shouldEqual s"Folder '${storage.folder}' does not exists on the endpoint '${storage.endpoint}'"
+    }
+
+    "fetch file" in new Ctx {
+      val source = sourceInChunks(genString())
+      client.getFile(storage.folder, path) shouldReturn IO(source)
+      val fetch = new ExternalDiskStorageOperations.Fetch[IO](storage, client)
+      fetch.apply(attributes).ioValue
```
hygt: It wouldn't hurt to check that the source and output are the same string

src/main/scala/ch/epfl/bluebrain/nexus/kg/storage/Storage.scala
```diff
@@ -297,6 +341,39 @@ object Storage {
     // format: on
   }
 
+  private val folderRegex = "^[a-zA-Z0-9_-]*$"
```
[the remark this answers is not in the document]
umbreak: We could skip validation totally, and leave it for the `storageClient.exists(name)` method. I think it is the external storage service responsibility.

src/test/scala/ch/epfl/bluebrain/nexus/kg/storage/ExternalDiskStorageOperationsSpec.scala
umbreak: I wasn't aware that unicode arrows were a problem. If I "fix" this, I do it everywhere

hygt: It's bad practice since ligatures are supported by IDEs, we don't really see it anywhere outside of Akka's codebase.

PR #99, opened 2017-11-10
Indexing ACLs - Add Kafka consumer

hygt:

PR #655, opened 2019-05-09
Allow listings with more than 10k entries

wwajerowicz:
Fixes https://github.com/BlueBrain/nexus/issues/99

### DIFF
--- a/src/main/scala/ch/epfl/bluebrain/nexus/kg/search/QueryResultEncoder.scala
+++ b/src/main/scala/ch/epfl/bluebrain/nexus/kg/search/QueryResultEncoder.scala
@@ -1,5 +1,7 @@
 package ch.epfl.bluebrain.nexus.kg.search
 
+import akka.http.scaladsl.model.Uri
+import akka.http.scaladsl.model.Uri.Query
 import cats.Id
 import cats.instances.either._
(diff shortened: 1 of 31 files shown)


src/main/scala/ch/epfl/bluebrain/nexus/kg/search/QueryResultEncoder.scala
```diff
@@ -48,22 +52,35 @@ object QueryResultEncoder {
...
+                .toMap + (QueryDirectives.searchAfter -> s.asJson.noSpaces) - QueryDirectives.from)
+
+          results deepMerge Json.obj(nxv.next.prefix -> Json.fromString(searchUri.withQuery(nextQuery).toString()))
+        case None => results
+      }
+    }
     Encoder.instance {
       case UnscoredQueryResults(total, list) =>
-        json(total, list)
+        json(total, list, list.lastOption.flatMap(_.sort))
```
umbreak: That seems hacky. Maybe we should have a `sort` also in the `QueryResults`?
At the same time, that sort is `Option[Seq[Json]]` is weird to me. Shouldn't Sort be [SortList](https://github.com/BlueBrain/nexus-commons/blob/master/modules/core/src/main/scala/ch/epfl/bluebrain/nexus/commons/search/SortList.scala)??

PR #671, opened 2019-05-29
Displayed all the property paths on incoming and outgoing

umbreak:


src/main/resources/blazegraph/incoming.txt
```diff
@@ -28,8 +36,6 @@ WHERE {
   {
     SELECT *
     WHERE { INCLUDE %resultSet }
-    ORDER BY ?s ?property
-    LIMIT {size}
-    OFFSET {offset}
-   }
+    ORDER BY ?_createdBy ?s ?paths LIMIT 20 OFFSET 0
```
bogdanromanx: limit and offset template?

src/main/resources/blazegraph/outgoing_include_external.txt
```diff
@@ -1,46 +1,50 @@
...
-}
\ No newline at end of file
+    SELECT (COUNT(?s) AS ?total)
+    WHERE { INCLUDE %resultSet }
+  }
+  UNION
+  {
+    SELECT *
+    WHERE { INCLUDE %resultSet }
+    ORDER BY ?_createdBy ?s ?paths LIMIT 20 OFFSET 0
```
bogdanromanx: limit and offset template?

src/main/resources/blazegraph/outgoing_scoped.txt
```diff
@@ -1,40 +1,44 @@
...
-}
\ No newline at end of file
+    SELECT (COUNT(?s) AS ?total)
+    WHERE { INCLUDE %resultSet }
+  }
+  UNION
+  {
+    SELECT *
+    WHERE { INCLUDE %resultSet }
+    ORDER BY ?_createdBy ?s ?paths LIMIT 20 OFFSET 0
```
bogdanromanx: limit and offset template?

umbreak:
Of course I have run them. Most of the part of the query is happening inside the scope of a resource, where there are very few triples.

The part which is not happening on the scope of the graph is after already selecting the links, which shouldn't be excessive. A regular payload is not supposed to have 1000 of links.

So I think it if more than good enough.

PR #676, opened 2019-06-07
Fix circular dependencies in context resolution

wwajerowicz:


src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Resources.scala
```diff
@@ -418,28 +420,42 @@ class Resources[F[_]](implicit F: MonadError[F, Throwable],
     lookup(Map.empty, importsValues(resId.value, graph).toList)
   }
 
-  private def materialize(projectRef: ProjectRef, schema: Ref, source: Json)(
+  private def materialize(projectRef: ProjectRef, schema: Ref, source: Json, sourceRef: Option[Ref])(
       implicit project: Project): EitherT[F, Rejection, ResourceF.Value] = {
 
-    def flattenCtx(refs: List[Ref], contextValue: Json): EitherT[F, Rejection, Json] =
+    def rejectionOrUnit(dependencyGraph: G[Ref, LkDiEdge]): EitherT[F, Rejection, Unit] =
```
umbreak: I think there is no need if the dependencyGraph, just with the previous `List[Ref]` should work

PR #686, opened 2019-07-02
Fixed issues related with empty tags on resources

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/684
Fixes https://github.com/BlueBrain/nexus/issues/685

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -1,4 +1,4 @@
 resolvers += Resolver.bintrayRepo("bbp", "nexus-releases")
 
-addSbtPlugin("ch.epfl.bluebrain.nexus" % "sbt-nexus"     % "0.11.0")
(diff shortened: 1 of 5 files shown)


project/plugins.sbt
```diff
@@ -1,4 +1,4 @@
 resolvers += Resolver.bintrayRepo("bbp", "nexus-releases")
 
-addSbtPlugin("ch.epfl.bluebrain.nexus" % "sbt-nexus"     % "0.10.13")
+addSbtPlugin("ch.epfl.bluebrain.nexus" % "sbt-nexus"     % "0.11.0")
```
bogdanromanx: Don't bump to this version yet, it drops the agents for monitoring which is not yet supported in KG.

PR #698, opened 2019-07-17
Prevented from empty filename and path on linking files payload

umbreak:


src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/package.scala
```diff
@@ -129,4 +131,7 @@ package object resources {
       .retry
   }
 
+  private[resources] def nonEmpty(s: String): EncoderResult[String] =
+    if (s.trim.isEmpty) Left(IllegalConversion("Field cannot be empty")) else Right(s)
```
wwajerowicz: We should also add the name of the field which is empty, otherwise the error might be quite confusing.

PR #703, opened 2019-07-24
Added async process to compute file digest

umbreak:
- Added a `UpdateFileDigest` command
- Added a `FileDigestUpdated` event
- Added an async process which reads events by tag (project={projectUuid}) and checks which files require a digest update. For those, it asks the StorageOperations to fetch the digest. There is a retry mechanisms in place to attempt to fetch the digest if the previously fetched digest was empty
- Added routes to manually apply a digest update: PATCH `/v1/files/{org}/{proj}/{file}`

### DIFF
--- a/src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Files.scala
+++ b/src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Files.scala
@@ -73,9 +73,9 @@ class Files[F[_]: Effect: Timer](repo: Repo[F])(implicit storageCache: StorageCa
       currFile          <- EitherT.fromEither[F](curr.file.toRight(notFound(id.ref)))
       (storageRef, attr) = currFile
       storage           <- EitherT.fromOptionF(storageCache.get(id.parent, storageRef.id), UnexpectedState(storageRef.id.ref))
-      digest            <- if(attr.digest == Digest.empty) EitherT.right(storage.fetchDigest.apply(attr.path)) else EitherT.leftT[F, Digest](ResourceAlreadyExists(id.ref): Rejection)
+      digest            <- if(attr.digest == Digest.empty) EitherT.right(storage.fetchDigest.apply(attr.path)) else EitherT.leftT[F, Digest](FileDigestAlreadyExists(id.ref): Rejection)
       _                 <- if(digest == Digest.empty) EitherT.leftT[F, Resource](FileDigestNotComputed(id.ref)) else EitherT.rightT[F, Rejection](())
-      updated           <- repo.updateFileDigest(id, storage.reference, curr.rev, digest)
+      updated           <- repo.updateDigest(id, storage.reference, curr.rev, digest)
     } yield updated
   // format: on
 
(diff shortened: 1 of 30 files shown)


src/main/scala/ch/epfl/bluebrain/nexus/kg/resources/Files.scala
```diff
@@ -59,6 +59,38 @@ class Files[F[_]: Effect: Timer](repo: Repo[F])(implicit storageCache: StorageCa
...
+  def updateDigestIfEmpty(id: ResId)(implicit subject: Subject, fetchDigest: FetchDigest[F]): RejOrResource[F] =
+    // format: off
+    for {
+      curr              <- repo.get(id, Some(fileRef)).toRight(notFound(id.ref))
+      currFile          <- EitherT.fromEither[F](curr.file.toRight(notFound(id.ref)))
+      (storageRef, attr) = currFile
+      storage           <- EitherT.fromOptionF(storageCache.get(id.parent, storageRef.id), UnexpectedState(storageRef.id.ref))
+      digest            <- if(attr.digest == Digest.empty) EitherT.right(storage.fetchDigest.apply(attr.path)) else EitherT.leftT[F, Digest](ResourceAlreadyExists(id.ref): Rejection)
+      _                 <- if(digest == Digest.empty) EitherT.leftT[F, Resource](FileDigestNotComputed(id.ref)) else EitherT.rightT[F, Rejection](())
+      updated           <- repo.updateFileDigest(id, storage.reference, curr.rev, digest)
```
bogdanromanx: `updated <- updateDigest(...)`?

src/main/scala/ch/epfl/bluebrain/nexus/kg/storage/FileDigestProjection.scala
```diff
@@ -0,0 +1,78 @@
...
+    val ignoreIndex: List[Resource] => F[Unit] = _ => F.unit
+
+    TagProjection.start(
+      ProjectionConfig
+        .builder[F]
+        .name(s"digest-computation-${project.uuid}")
+        .tag(s"project=${project.uuid}")
+        .plugin(config.persistence.queryJournalPlugin)
+        .retry[KgError](config.storage.digestRetry.retryStrategy)(kgErrorMonadError)
+        .batch(1, 1 millis)
```
bogdanromanx: we should store these values in the config

src/main/scala/ch/epfl/bluebrain/nexus/kg/storage/FileDigestProjection.scala
```diff
@@ -0,0 +1,78 @@
...
+    P: Projections[F, Event],
+    as: ActorSystem,
+    config: AppConfig,
+    F: Effect[F]): StreamSupervisor[F, ProjectionProgress] = {
+
+    val mapper = new FileDigestProjectionMapping(files)(fetchDigest, kgErrorMonadError)
+
+    val ignoreIndex: List[Resource] => F[Unit] = _ => F.unit
+
+    TagProjection.start(
```
bogdanromanx: please also add the Kamon statistics part to the projection, just like in ElasticSearchIndexer so it can be exposed to Prometheus

src/main/scala/ch/epfl/bluebrain/nexus/kg/storage/FileDigestProjection.scala
umbreak: This does not make sense, since the batching here is doing nothing. Actually I'm going to remove the batch line.

PR #718, opened 2019-08-21
Added ProjectDigestCoordinator

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/730


src/main/scala/ch/epfl/bluebrain/nexus/kg/async/ProjectViewCoordinator.scala
```diff
@@ -31,26 +31,17 @@ import monix.execution.Scheduler.Implicits.global
...
-          processedEvents = p.processedEvents,
-          discardedEvents = p.discardedEvents,
-          totalEvents = p.totalEvents,
-          lastEventDateTime = p.lastEvent,
-          lastProcessedEventDateTime = p.lastProcessedEvent
-        )
-    }
-  }
+  def statistics(project: Project, view: SingleView): F[Statistics] =
+    IO.fromFuture(IO(ref ? FetchProgress(project.uuid, view))).to[F].map { case s: Statistics => s }
```
bogdanromanx: same comment on the match error

src/main/scala/ch/epfl/bluebrain/nexus/kg/async/ProjectDigestCoordinatorActor.scala
```diff
@@ -0,0 +1,206 @@
...
+
+  def startCoordinator(project: Project, restartOffset: Boolean): StreamSupervisor[Task, ProjectionProgress]
+
+  private def progress: Task[Statistics] =
+    child.map(_.state().map(_.getOrElse(NoProgress))).getOrElse(Task.pure(NoProgress)).map { p =>
+      Statistics(p.processedCount, p.discardedCount, p.processedCount, p.offset.asInstant, p.offset.asInstant)
+    }
+}
+
+private class FileDigestEventMapping[F[_]: FetchDigest](files: Files[F])(implicit F: MonadError[F, KgError]) {
```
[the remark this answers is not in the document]
umbreak: Because as you can see I only use this MonadError to throw errors of subtype `KgError`. And because the TagProjections will only retry for those...so the more concise the better, right?

src/test/resources/resolve/default_elasticsearch_pipeline.json
```diff
@@ -0,0 +1,76 @@
+{
```
bogdanromanx: these files should not be here

PR #738, opened 2019-11-01
Added support for Composite View (initial iteration)

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/747


src/main/scala/ch/epfl/bluebrain/nexus/kg/routes/ViewRoutes.scala
```diff
@@ -208,58 +208,93 @@ class ViewRoutes private[routes] (
...
+          hasPermission(query).apply {
+            entity(as[String]) { query =>
+              val result = viewCache.getBy[View](project.ref, id).flatMap(runSearch(id, query))
+              complete(result.runWithStatus(StatusCodes.OK))
+            }
+          }
+        }
+      },
+      (post & pathPrefix(IdSegment / IdSegment / "sparql") & pathEndOrSingleSlash) { (id, projId) =>
+        operationName(s"/${config.http.prefix}/views/{}/{}/{}/{}/sparql") {
```
bogdanromanx: how about changing the APIs to this:
`/{prefix}/views/{org}/{proj}/{view_id}/projections/{projection_id}/*`

PR #745, opened 2019-12-02
Added interval restart

umbreak:
Fixes https://github.com/BlueBrain/nexus/issues/919

This is the first iteration of the Interval feature.

TODO:
- Only restart when new events are present in the project (compared to the previous run)
- Calculate the interval relatively to the previous run (for that we need to store the previous restart time on the progress). This will be useful on actor rebalancing and on service restarts


src/main/scala/ch/epfl/bluebrain/nexus/kg/async/ProjectViewCoordinatorActor.scala
```diff
@@ -156,92 +162,104 @@ private abstract class ProjectViewCoordinatorActor(viewCache: ViewCache[Task])(
...
+            stopView(view, coordinator, deleteIndices = false)
               .map(_ => self ! ViewsAddedOrModified(project.uuid, restartOffset = true, Set(view)))
+          case None => sender() ! None
+        }
+
+      case RestartProjections(uuid, viewId) =>
+        val _ = children.findBy[CompositeView](viewId) match {
+          case Some((view, coordinator)) =>
+            sender() ! Option(Ack(uuid))
+            logStop(view, "restart triggered from client")
```
wwajerowicz: We should probably log differently when a restart was caused by scheduled and when it was caused by client. It will probably make the debugging a bit easier.

umbreak: Yeah, I'll add something to this logging in the next PR

